Issue #33, opened 2019-03-20
Group update


MrLarssonJr:
Me (Jesper Larsson, jespel@kth.se) and Benjamin Tellström (@btellstrom btel@kth.se) voluntereed to do a presentation for next monday. So we are gonna need to be group for that.

Our preassinged groups were T1G15 and T1G25.

SophieHYe:
#37 merged

PR #37, opened 2019-03-20
Created presentation task

btellstrom:
The group number may be preliminary, but it is the one I was assigned to. Read more in #33 .


SophieHYe:
Hi, No problem. After merging your request, I will update the folder name to jespel-btel as we are dropping the number of groups now.

PR #43, opened 2019-03-20
presentation OCI

Sjobom:
Anders Sjöbom asjobom@kth.se
Valtteri Lehtinen vleh@kth.se

We would like to to do a presentation about the Open Container Initiative.
It's an organization founded by Docker, CoreOS and other leaders in the container industry to create a formal standard for containers.
www.opencontainers.org.

We would like to present on the 1 of april since that lecture aligns with the topic.


SophieHYe:
After merging your request, I will update the folder name to asjobom-vleh as we are dropping the number of groups now.

PR #46, opened 2019-03-20
Added a presentation proposal

hannesrabo:


juliuscc:
We specifially want to present during the micro service lecture

SophieHYe:
After merging your request, I will update the folder name to hrabo-jcelik as we are dropping the number of groups now.

PR #48, opened 2019-03-20
fix typo in readme

EmilGedda:
As of d1bfb1c there are only 5 categories now.
This commit simply changes "six" to "five" in one sentence under the heading Group Rules.


SophieHYe:
THANKS!

PR #58, opened 2019-03-21
correct presentation time

Sjobom:
In #43 we asked to present on the 1st of april (week 3) since it will fit the topic of the lecture better. We are talking about the Open Container Intiative and the lecture in week 3 is about containers etc.


SophieHYe:
Thanks for your correction!

PR #60, opened 2019-03-21
Moved presentation proposal to the correct week

hannesrabo:
The presentation was placed on the wrong week in regards to topic covered. Moved it to the correct week.


SophieHYe:
Great! Thanks for your correction!

PR #70, opened 2019-03-25
Presentation week 2: Are automatically generated test suites "good"?

slarse:
Hello! This is our presentation for today. We changed the title from 'What makes a good test suite?' to 'Are automatically generated test suites "good"?' as discussing automatically generated tests ended up being the point of the presentation.

The content of the presentation is still as described in the proposal, but weights shifted a bit over the weekend!

Also updates the agenda with the new title.


gluckzhang:
Merged, thanks!

PR #87, opened 2019-03-27
Essay: Graph Databases

slarse:
Hello!

The instructions said we could do a task solo, so I'd like to do that for this essay. I have partners for all other tasks.

I'd like to write an essay on graph databases. I don't know much about them at this time, but I intuitively feel like they are relevant to DevOps due to being schemaless (rapid iteration) and being capable of modeling complex relationships (for example useful to keep track of infrastructure that is constantly in flux).

I will make comparisons with relational databases, and possibly other NoSQL databases (e.g. doc dbs) if there is space left in the essay.


gluckzhang:
Hi @slarse , no problem, the proposal sounds interesting, thanks :)

PR #109, opened 2019-04-02
Add essay proposal

jaokob:
We want to write about wasm and the future of containers, see readme.


gluckzhang:
Nice topic, thanks @jaokob

PR #113, opened 2019-04-02
added proposal for presentation

protopapa:


gluckzhang:
Hi @protopapa , the presentation topic sounds great. Regarding your open task, will you be able to demostrate it on our Michelin Devops Demo Day, May 6th? Because it sounds like a demo task as well, I'm wondering which category we should put it into.

protopapa:
For demo probably we (with the classmate I found) will for for ELK and how it can help developers. 
Another open task could be how to go from Microservice to micro-microservices and what infra is needed. But is a bit big task.

gluckzhang:
Cool, since you have already got a good plan for demo, I'll merge this PR then :)
Thanks!

PR #120, opened 2019-04-04
Add essay proposal README

Eika96:
# Pipelines

* A comparison between the more *traditional* CI/CD and the more modern pipeline approach.
* Use cases of pipelines.
* Discussion about the advantages/disadvantages of modern pipelines.
* Available technologies and their differences.


gluckzhang:
Registered, thanks @Eika96 !

PR #121, opened 2019-04-04
demo proposal: GitLab Auto Devops and GKE

Sjobom:
I want to set up a CD Pipeline with GitLab's Auto Devops and automatically deploy code to a Kubernetes Cluster on GKE.

The best would be if I can find someone that wants to do this together with me. But If I can't I would like to do it by myself.

If the PR is accepted I will search for someone to do this with via an issue.


gluckzhang:
Hi @Sjobom , the proposal sounds great, it's accepted :)

PR #125, opened 2019-04-05
Move presentation to correct folder

felixkollin:
Noticed that the presentation should be placed inside the week 4 folder, not the presentations root.

Sorry for this.


gluckzhang:
Thank you, I didn't notice this just now. Have a nice weekend!

PR #128, opened 2019-04-08
Demo for Haris Adzemovic (harisa)

HarisAdzemovic:
I'd like to demo how one can use statical analysis
and automatic repair to fix some errors without user
input.


gluckzhang:
As one of the key contributors in Spoon project, Martin will like your demo :)

PR #137, opened 2019-04-08
 Register tonik-krijor for competition

tonikarppi:


gluckzhang:
Thanks for the update!

PR #140, opened 2019-04-09
Demo proposal

gustafguner:


gluckzhang:
Merged, thanks @gustafguner . And note that if you want to automatically deploy your application in K8s, maybe you also need a CD tool like Jenkins.

PR #143, opened 2019-04-11
Add essay proposal

ghodt:
#  A/B-testing

## Members
- Emma Good, egood@kth.se

## Topic
I want to write about A/B-testing: what it is, why it's used, and look at some cases where A/B-testing has been used.


gluckzhang:
Merged, thanks @ghodt :)

PR #145, opened 2019-04-11
Open proposal

jaokob:
An open proposal. Will this suffice or is more needed?


gluckzhang:
Hi @jaokob , I think it's a good proposal. Thanks :)

PR #146, opened 2019-04-12
Add lcjna essay topic

LouisITB:
## Essay topic

Why you should be using DevOps.

An example based article on how the tasks you are doing can be automated, a short introduction of DevOps and the history behind its creation with industry examples to back it up.

## Members

- Louis Nathan, lcjna@kth.se


gluckzhang:
Merged, thanks @LouisITB :)

PR #148, opened 2019-04-13
Add open proposal for aeri3 and egedda: Fuzzing of JSON parsing libraries

Anders-E:
Proposal for open assignment with @EmilGedda.


gluckzhang:
Interesting topic with a nice proposal, thanks @Anders-E and @EmilGedda !

PR #158, opened 2019-04-17
Add essay proposal

jqhan:


gluckzhang:
Good topic, thanks @jqhan !

PR #162, opened 2019-04-18
Proposal for essay added, containers and serverless

oscros:


gluckzhang:
The topic sounds nice, thanks @oscros .

PR #165, opened 2019-04-22
Open PR 

jaokob:
We wish to merge our open contribution. We could also add the code and the dockerfile for the nodeJS backend we are using, but it is so basic that we did not see any need (three endpoints essentially). Everything that is needed to launch the stack is already present.


gluckzhang:
Nice work, thanks @jaokob and @adibbin !

adibbin:
@gluckzhang Does this mean we have passed or are we waiting for a result?

gluckzhang:
> @gluckzhang Does this mean we have passed or are we waiting for a result?

@adibbin , I just merged the work first, then we will evaluate the work systematically. We will notice you the result later :)

adibbin:
Okay, thanks for the fast reply! 😅

PR #182, opened 2019-04-24
[Demo] Add Git plumbing demo proposal

slarse:
Hello! Please see the README for our demo proposal.

We are a bit unsure of the topic as it is more on the small scale of building blocks rather than the grand multi-server cloud [many buzzwords omitted] kind of thing that the instructions allude to, but we still think it's an interesting and absolutely fundamental topic for DevOps.


gluckzhang:
Hi @slarse , I agree with you. The topic is very nice and helpful, the PR has been merged :)

PR #194, opened 2019-04-26
Add Open Task proposal

fredrikfloven:


gluckzhang:
Thanks @fredrikfloven . This sounds like a Chrome extension for GitHub, which is interesting and useful :)

PR #195, opened 2019-04-27
add essay proposal

yipeitu:
**Members**
Yi-Pei Tu (yptu@kth.se)

**Topic**
 The relation between load balancing and auto-scaling

**Structure**
-Why we need load balancing and auto-scaling?
-What is load balancing?
-What is auto-scaling?
-The relation between both of them.
-What should address/evaluate in practical case?


gluckzhang:
Interesting topic, thanks @yipeitu .

PR #196, opened 2019-04-27
[Essay] An Introduction to Graph Databases

slarse:
I finally managed to finish my essay. I initially wanted to also include RDF graphs in the practical usage comparison, but the whole thing just kind of ballooned in size so I abandoned that idea. I still found the perspective valuable, so I kept the background info on RDF.

I'm hosting the HTML version of the essay [on my website](https://slar.se/essays/graph_db_essay.html), and a PDF version is included in the repo. Getting the markdown source to compile nicely both to PDF (through LaTeX) and to HTML was quite the adventure.


gluckzhang:
Hi @slarse , really nice work, thank you!

PR #209, opened 2019-04-30
Essay submission

netrounds-erik:
Here is our essay submission.
Added it as a pdf in our essay folder.


gluckzhang:
Got it, thanks @netrounds-erik

PR #225, opened 2019-04-30
Adds pdf of essay

btellstrom:
My essay.


gluckzhang:
Good job, thanks @btellstrom

PR #228, opened 2019-04-30
Essay asjobom-krijor

Sjobom:
Submission for our (asjobom-krijor) essay!


gluckzhang:
Good job, thanks @Sjobom !

PR #237, opened 2019-04-30
Submit Essay pdf

axelminn:
here is my essay in a pdf


gluckzhang:
Merged, thanks @axelminn

PR #244, opened 2019-05-01
Add essay submission

jqhan:
Since my first submission was late anyway, I decided to improve it because it felt rushed. I understand that I can't get a higher grade than pass due to it being late, but it will at least be a better read for ppl.


gluckzhang:
Got that, I'll merge it first. Thanks @jqhan .

Issue #18, opened 2018-10-24
DevOps and Security


monperrus:
Security is important in the context of DevOps. What are good introductory pointers about DevOps and Security?

(Credits: @sbuc)

monperrus:
Runtime application self-protection: https://en.wikipedia.org/wiki/Runtime_application_self-protection

monperrus:
Principles:
* Complete Mediation Principle (useful for APIs)
* Least privileged

monperrus:
Intrusion detection: https://en.wikipedia.org/wiki/Intrusion_detection_system

(signature based, anomaly detection)

monperrus:
Very good set of pointers: https://www.sqreen.io/checklists/devops-security-checklist

sbuc:
Mapping security design principles to devops on one axis, mapping security concepts/mechanisms to devops on another.

lsc:
Dynamic and short lived secrets for authorisation, see for example AWS IAM Roles are implemented or [Hashicorp Vault](https://www.vaultproject.io/).

monperrus:
Open Source: Simplifying Serverless Secrets
https://open.nytimes.com/open-source-simplifying-serverless-secrets-in-google-cloud-a95451e545b1

bbaudry:
Vault and kubernetes
https://github.com/kelseyhightower/vault-on-google-kubernetes-engine

monperrus:
CI/CD enables automated program hardening:

Operating system protection through program evolution, Fred Cohen, 1993

monperrus:
Check your repos... Crypto-coin-stealing code sneaks into fairly popular NPM lib (2m downloads per week)
https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/

monperrus:
7 Tips for Container and Kubernetes Security
http://lxer.com/module/newswire/ext_link.php?rid=264809

gluckzhang:
[Microservices Hierarchy of Needs](https://medium.com/@bibryam/microservices-hierarchy-of-needs-151df49fc678)
[KUBERNETES: AN OVERVIEW](https://thenewstack.io/kubernetes-an-overview/) (This is a nice introduction to Kubernetes architecture and advantages)

monperrus:
On The Relation Between Outdated Docker Containers, Severity Vulnerabilities and Bugs.
http://arxiv.org/abs/1811.12874

bbaudry:
[...]

PR #253, opened 2019-05-03
add demo suggestion

fredrikfloven:


gluckzhang:
Merged, please note that your group number is #18 and you will present the demo in session 09:15-10:00.
See you next Monday!

Issue #19, opened 2018-10-24
Information model


AG00:
There is a relationship between desired properties (requirements), tests (verification) and test results. Those may be present in a hierarchy of components/subsystems/libraries. In addition, there is a relationship between planning activities on the system (product) being developed and its desired properties. Further, with time there will be more than one system/product configuration to consider and more than one version in use with different installations/deployments/customers. These relationships quickly becomes complex to understand and keep track of. As changes are being made (code or documents) and automatic tests are run, it would be good to have the information model changes generated rather than manually written. The information model can also be used to convince a customer (or the devops team itself) that a complex/high-level system property is achieved.

monperrus:
Thanks, merged with #9

PR #254, opened 2019-05-03
Add demo proposal

DokID:


gluckzhang:
Merged, please note that your group number is `#19` and you will present the demo in session 10:15-11:00.
See you next Monday!

Issue #20, opened 2018-10-24
Devops-team habits


AG00:
Each organisation has a culture. A team has its own way of working within a larger organisation. Even if there are formal process descriptions, written rules and guidelines, I believe the team's own habits affects the result much more than anything else. It is, in a sense, a shared memory of how things are done. Especially if the team members are co-located and a bit isolated/protected from the rest of the organisation.  The question is: how to develop the right habits in a devops team? 

Litterature suggestion: "The power of habit - Why we do what we do in life and business" by Charles Duhigg.

monperrus:
I propose to merge this issue with #7. Could you post your comment there and close this one? Thanks!

PR #256, opened 2019-05-04
Add Demo Topic 

NicoleCarter12:


gluckzhang:
Merged, please note that your group number is `#20` and you will present the demo in session 09:15-10:00.
See you next Monday!

Issue #22, opened 2018-11-06
Infrastructure as code


alanmcg:
Wikipedia page: https://en.wikipedia.org/wiki/Infrastructure_as_Code

Original comment:
As mentioned by someone in another ticket, infrastructure as code was the "next big thing" not long ago (it still is a big thing, but it's already here).

Terraform is the leader in this area with competitors such as AWS cloud formation.

"infra as code" comes with it's own issues, you can just as easily automate the total destruction of your production environment as you can create a new one if you are not careful. There is a lot of documentation and blog posts on using terraform etc, but they are typically a bit out of date. A devops course with a section on IAC would be awesome

monperrus:
DNS (https://en.wikipedia.org/wiki/Domain_Name_System) is a key DevOps infrastructure (and that's the code service of Cloudflare)

monperrus:
A Systematic Mapping Study of Infrastructure as Code Research
https://www.sciencedirect.com/science/article/pii/S0950584918302507

bbaudry:
Terraform
https://www.terraform.io/

bbaudry:
Infra as code
https://itnext.io/docker-makefile-x-ops-sharing-infra-as-code-parts-ea6fa0d22946

monperrus:
Pets versus Cattle

Stop Killing Your Cattle: Server Infrastructure Advice
https://www.linuxjournal.com/content/stop-killing-your-cattle

bbaudry:
Security Smells in Infrastructure as Code Scripts

https://arxiv.org/abs/1907.07159

monperrus:
What about merging this thread into #2?

monperrus:
Merged into #2

PR #258, opened 2019-05-04
demo proposal 

axelminn:
Introduction to Snort - DevOps tool for security


gluckzhang:
Merged, please note that your group number is `#22` and you will present the demo in session 09:15-10:00.
See you next Monday!

PR #259, opened 2019-05-05
add Felix as member and update demo description

Sjobom:


gluckzhang:
Got it, your group number is still `#2`, see you tomorrow!

PR #270, opened 2019-05-05
Add link to demo screencast

ghodt:
And remove member from demo group.


gluckzhang:
Thanks for the contribution!

PR #272, opened 2019-05-05
Add Demo

arnthor89:


gluckzhang:
Thanks for the contribution!

gluckzhang:
Hi @arnthor89 , just for confirmation, have you submitted the screencast already? I didn't find the link to your video in this submission.

arnthor89:
Hi @gluckzhang, if you look under the project I have two directories 1) aws 2) k8s there is a screencast in both of them. Did mean to upload only one of them but then I figured I might as well upload both since they were already made and someone might find it useful. However, I only demoed aws during the demo day so maybe a should include that in the readme in the top directory?

PR #275, opened 2019-05-06
Add screencast URL

netrounds-erik:
Add the url to the screencast of the demo


gluckzhang:
Thanks for the contribution!

PR #283, opened 2019-05-10
Open assignment update

Anders-E:
The goal of our open assignment was to contribute to open-source JSON-libraries. The issue we created as a result of our open assignment has now been resolved with our bug-causing JSON-input added as test cases to the project. As such, we would like this to show in our README even if it happened after the submission deadline.


gluckzhang:
That's excellent, congrats!

PR #290, opened 2019-05-22
Added demo screencast

DokID:


DokID:
Added essay resubmission @gluckzhang @bbaudry

PR #296, opened 2019-05-29
Update Essay

mudaliarkartik:
Updated the essay based on feedback received.

Regards
- Kartik Kandavel Mudaliar(mudaliar@kth.se)
- Henry Luong Phu(talip@kth.se)

PR #300, opened 2019-06-01
Feat: Update essay

Tailp:
Submit the revised version of essay (duplicated PR #296 before deadline).


mudaliarkartik:
Based on #296

PR #331, opened 2020-03-17
Presentation Topic for week 2

gynther-k:
Pull request for Kasper Liu and Gunnar Applelid


cesarsotovalero:
Thanks @gynther-k, it is a good topic for presentation :)

PR #334, opened 2020-03-17
Add initial essay proposal

ruwaid4:
# Essay Proposal
## Members
Ruwaid Louis (ruwaid@kth.se)
GitHub Username: ruwaid4

## Topic
The monitoring process:
* What tools are used
* What is monitored for different applications
* How the monitored data is used
* What some standard practices are

### DIFF
--- a/contributions/essay/ruwaid/README.md
+++ b/contributions/essay/ruwaid/README.md
@@ -1,0 +1,11 @@
+# Essay Proposal
+## Members
+Ruwaid Louis (ruwaid@kth.se)
+GitHub Username: ruwaid4
+
+## Topic
+The monitoring process:
+* What tools are used
+* What is monitored for different applications
+* How the monitored data is used
+* What some standard practices are


gluckzhang:
Sounds nice, thanks @ruwaid123 !

gluckzhang:
@ruwaid123 , to avoid the risk of being superficial or vague, you could think more about the scope and novelty even though the PR has been merged. For example, what are the 3 most related resources on the web and what are the novelty and differences of your own essay with respect to them?
We could have a meeting during lab sessions if you need more help :)

ruwaid4:
There's nothing really groundbreaking happening in the field of monitoring right now, the chosen essay topic is more for me to get a better understanding of how the monitoring process works and not really to bring something new or extraordinary. It is to be a summary but still dive deeper than what most articles do. I could only find sources describing different parts of the monitoring process or summarizing the whole process very shallowly.

ruwaid4:
@gluckzhang What about specifically looking into the monitoring process in the video game industry? Find out what kind of data is specific to video games and how it is used etc. Would that be a better scope? If so i will update the README.

gluckzhang:
> @gluckzhang What about specifically looking into the monitoring process in the video game industry? Find out what kind of data is specific to video games and how it is used etc. Would that be a better scope? If so i will update the README.

Hi @ruwaid4 , I prefer this proposal because you will introduce the monitoring techniques generally first and then go deeper into a specific scenario. Let's update the README in this way. Please also add some key resources (e.g., 3 links) which you are going to read and cite. Thanks!

ruwaid4:
@gluckzhang I have now updated the README, let me know if there is still something missing.

gluckzhang:
Hi @ruwaid4 , thanks for the updates. This version is much better. Since this PR has been merged already, you could make another PR to update the existing `README` file :)

PR #340, opened 2020-03-17
Add initial demo proposal

Zodbigt:
**Demo Proposal**

**Members**
Tommy Samuelsson (tommysam@kth.se)
Stina Långström (stinalan@kth.se)

**Topic**
Data corruption

The idéa is to securly use fetched data in applications. So the application in it self can relay on that data it will use in different components won't crash the application. If a data set is demed corrupted it will discard it then notfiy ennd user that data with ID:# was not renderd. (Notification can be via slack, mail or similar).

**Demo**
Two live pages the receive the same data. But the one not using the "data corruption handler" will crash when they receive the corrupted data set. 
The demo could be interactive by having a input field(s) for data. Allowing the demo-viewer to try to crash the site's. (Hopefully only one will crash)

### DIFF
--- a/contributions/demo/tommysam-stinalan/README.md
+++ b/contributions/demo/tommysam-stinalan/README.md
@@ -1,0 +1,14 @@
+# Demo Proposal
+## Members
+Tommy Samuelsson (tommysam@kth.se)
+Stina Långström (stinalan@kth.se)
+
+## Topic
+Data corruption
+
+The idéa is to securly use fetched data in applications. So the application in it self can relay on that data it will use in different components won't crash the application. If a data set is demed corrupted it will discard it then notfiy ennd user that data with ID:# was not renderd. (Notification can be via slack, mail or similar).
+
+
+## Demo
+Two live pages the receive the same data. But the one not using the "data corruption handler" will crash when they receive the corrupted data set. 
+The demo could be interactive by having a input field(s) for data. Allowing the demo-viewer to try to crash the site's. (Hopefully only one will crash)


stinalangstrom:
What do you think about this?

We are uncertain about how generic the solution must be, or can it be specified for a specific site/application?

Do you have any other suggestions that could improve our concept or give guidelines?

SophieHYe:
Hi,
Thanks for the contribution PR. I like your idea.
Some concern from me is about the scope of "data corruption handler".  
There could be different types of data corruption, are you to handle some specific types?
Another minor point is to  'notify end user ' about the data, I am not sure how many end users would be interested in that. But I think the developers instead would be more interested in the corrupted data. Maybe we can also think about 1) how to handle crash for end-users to achieve better usability and 2) record the corrupted data to reproduce the crash to help the development process.

Zodbigt:
Hi,
We wrote that part wrong. We meant of course for the developers managing the application/service.
Our first thought for the data corruption type was, for exaple a user input error. So for example the application/service expects an Integer but gets a string, or get empty or null values. 
So the "data corruption handler" would prevent the application from craching in that case.
Something in that direction at least.

stinalangstrom:
Here is our hosted demo: [link](https://dd2482-demo-data-corruption.firebaseapp.com)
If you add events that crashes the "regular site" you will experience a white page, to continue exploring the demo you can go back to main page and refresh that page. Since you have added events with bad types you should clear the database with the button on the main page.

PR #342, opened 2020-03-17
Presentation topic for week 3, Travis vs Jenkins, a comparison of two popular CI tools

jonasjohansson8908:
## Members
Jonas Johansson, jonasjo5@kth.se Github username: jonasjohansson8908
Marcus Jonsson Ewerbring, marcusew@kth.se Github username: Marcus9512

## Topic
A comparison that covers:
- How do they work, what differs?
- Different features?
- Design differences, why?
- How are they integrated with git? (Maybe setup an example)
- Pros and cons.

### DIFF
--- a/contributions/presentation/week3/jonasjo5-marcusew/README.md
+++ b/contributions/presentation/week3/jonasjo5-marcusew/README.md
@@ -4,7 +4,7 @@
 Jonas Johansson, jonasjo5@kth.se
 Github username: jonasjohansson8908
 
-Marcus Jonsson Ewerbring (maracusew@kth.se)
+Marcus Jonsson Ewerbring, marcusew@kth.se
 Github username: Marcus9512
 
 ## Topic
--- a/contributions/presentation/week3/jonasjo5-marcusew/README.md
+++ b/contributions/presentation/week3/jonasjo5-marcusew/README.md
@@ -8,3 +8,10 @@ Marcus Jonsson Ewerbring (maracusew@kth.se)
 Github username: Marcus9512
 
 ## Topic
+Travis vs Jenkins, a comparison of two popular CI tools.
+- How do they work, what differs?
+- Different features?
+- Design differences, why?
+- How are they integrated with git?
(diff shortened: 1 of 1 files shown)


SophieHYe:
Hi,
It's a very related and classical presentation. 
Could you please also consider adding one point to identify the original and reflective part of your representation compared to the existing comparisons?

jonasjohansson8908:
@SophieHYe 
We are thinking of adding one point where we reflect on our experience setting up each service. Neither of us have set up a CI server before so we think that this part could also be quite humorous when going through the points where we (probably) failed. 

Something like:

- 2 Noobs experience setting up continous integration servers.

Would this also count as original? We couldn't find anything similar when researching on the topic.

rarkins:
FYI I think a bigger trend in CI is "traditional" CI like Jenkins and Travis vs the more integrated approach of GitLab's Pipelines and GitHub's Actions.

Also, CircleCI is by far and away the leading "hosted" CI service so I would pick that over Travis if you're looking for a service that will be most relevant in 1-2 years time.

gluckzhang:
> @SophieHYe
> We are thinking of adding one point where we reflect on our experience setting up each service. Neither of us have set up a CI server before so we think that this part could also be quite humorous when going through the points where we (probably) failed.
> 
> Something like:
> 
> * 2 Noobs experience setting up continous integration servers.
> 
> Would this also count as original? We couldn't find anything similar when researching on the topic.

The idea is interesting. It also sounds like you will compare the onboarding experience of Travis and Jenkins, where nice features for entry-level users could be introduced :)

@SophieHYe , what do you think?

jonasjohansson8908:
@gluckzhang Thank you for the response and thank you @rarkins for the great suggestions. Will try to look into CircleCI but we are almost done with the onboarding experience and reflection of travis and jenkins. So if it's ok we will cover them in our presentation.

gluckzhang:
Hi @jonasjohansson8908 , no worries, I think it's totally fine :)

PR #346, opened 2020-03-17
Presentation week2 bus factor

Mema5:
Hi,

@valsen and I want to make a presentation next week around the "bus factor". 

We also want to ask the TA for some guidance. We found three papers about computing the "bus factor" for git projects. If we go into details it can make a full presentation. But we are not sure if it is the most interesting to only stick to that... We could extend to gender equality and diversity in git repo contributors, ask the open question "what is the profile of an open source contributor?" or else look into how we can assess the health of a repo by looking at the git history.

What's your views about that?

### DIFF
--- a/contributions/presentation/week2/madon-vals/README.md
+++ b/contributions/presentation/week2/madon-vals/README.md
@@ -11,9 +11,13 @@ Github username: Mema5
 Health of git repositories: the "bus factor". 
 
 ## Motivation
-An interesting measurement to assess the health of a software project in the "bus factor": the minimal number of developers that have to be hit by a bus before a project is incapacitated. Reasearchers have proposed algorithms to evaluate this number based on git history. 
+An interesting measurement to assess the health of a software project in the "bus factor": the minimal number of developers that have to be hit by a bus before a project is incapacitated. Researchers have proposed algorithms to evaluate this number based on git history. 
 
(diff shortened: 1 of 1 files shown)


SophieHYe:
Hello,
Thanks for bringing an interesting topic. I like the idea very much if I understand correctly.

Both the open question you proposed "what is the profile of an open source contributor? or else look into how we can assess the health of a repo by looking at the git history " are interesting. But
I personally would prefer the second question, as I assume more people care about the quality of repository than contributors. It could have much more impact on potential contributors and researchers. And considering the quality of open source repos has always been complained and the industry has a problem to trust and use open source projects. Assessing the health of a repo could bring good guidance to them. This could also be related to the open-source supply chain.

>> If we go into details it can make a full presentation.
I agree with you.  It could be an essay or even a paper to extend this topic.  Maybe you can also consider it :)
But for a presentation, you may consider comparing the three existing papers and present useful takeaways from them.
>>We could extend to gender equality and diversity in git repo contributors.
Good to extract features of contributors, maybe also consider the LOC commit, experience, familiar across the projects?

Mema5:
Hi @SophieHYe , thank you for your really interesting propositions! If we don't cover everything in the presentation, we will at least mention them in conclusion :)

We modified the README to precise our topic. Can you accept the pull request or do you need any more information?

PR #361, opened 2020-03-17
Week6 presentation: novel decentralized discussion protocol and tool

GustavDowling:
Excuse me for the late reply, I was working on some other things and also sketching out more concretely the features of the tool.

The idea behind the tool is like an internet forum, except everyone in the forum has a massive text file which also includes other information that can be parsed by the tool. The advantages of this are many, everyone can verify that previous posts have not been edited, everyone stores a backup and anyone can fork the forum. Additionally, it doesn't matter how the file is distributed, meaning that the discussion will not be lost in time as soon as a server goes down and no one needs to trust a server owner. The tool will also provide authentication of who wrote what using private/public key cryptography.

The use cases would be similar to any forum, although it has advantages when it comes to preserving older conversations.

Ok, the tool is pretty cool, but how does it relate to DevOps? I would say that the functionality is similar to git which is clearly a DevOps tool, but it is aimed at a more general audience. Therefore I think it falls under DevOps.


gluckzhang:
Hi @GustavDowling , thanks for proposing such an interesting topic. Could you please describe the outline of the presentation in bullet points? Something like:

- Sketching out the features of the protocol
- Giving use cases
- Introducing the tool
- ...

GustavDowling:
Excuse me for the late reply, I was working on some other things and also sketching out more concretely the features of the tool.

The idea behind the tool is like an internet forum, except everyone in the forum has a massive text file which also includes other information that can be parsed by the tool. The advantages of this are many, everyone can verify that previous posts have not been edited, everyone stores a backup and anyone can fork the forum. Additionally, it doesn't matter how the file is distributed, meaning that the discussion will not be lost in time as soon as a server goes down and no one needs to trust a server owner. The tool will also provide authentication of who wrote what using private/public key cryptography.

The use cases would be similar to any forum, although it has advantages when it comes to preserving older conversations.

Ok, the tool is pretty cool, but how does it relate to DevOps? I would say that the functionality is similar to git which is clearly a DevOps tool, but it is aimed at a more general audience. Therefore I think it falls under DevOps.

gluckzhang:
Hi @GustavDowling , thanks for the explanation, which helps a lot in understanding your idea. Could you add your answer in the `README`? Regarding my previous question, actually I mean how you are going to organize your presentation (an outline in bullet points).

An example here: https://github.com/KTH/devops-course/tree/master/contributions/presentation/week2/javierro-sasig#presentation-outline

GustavDowling:
Updated README, also, is it possible to move the presentation to week 5 or 6? More time would be nice so the presentation is not rushed.

gluckzhang:
Hi @GustavDowling , yes, you could push updates to this PR to change your task folder (both week 5 and 6 are fine). I will merge it after your update :)

GustavDowling:
Can you merge it now? Thank you for your help in organizing this

gluckzhang:
Merged. Thanks @GustavDowling :)

PR #363, opened 2020-03-17
The proposal for Tutorial

ashgreyship:
# Tutorial proposal: [gocd](https://www.gocd.org/pipelines-as-code.html)
##  Group members

 - Dingli Mao dingli@kth.se
 - Akshay Sinha akshays@kth.se

## Decsription
Gocd is a free and open-source Continuous Integration and Continuous Delivery system. We will provide all essential steps regarding install and set up Gocd server. In addition, we will examine and explain most functions of Gocd.

### DIFF
--- a/contributions/essay/dingli/README.md
+++ b/contributions/essay/dingli/README.md
@@ -1,0 +1,21 @@
+# The advantages and challenges while using Continuous Delivery
+## Members
+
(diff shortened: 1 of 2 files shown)


cesarsotovalero:
LGTM
Please always keep in mind the novelty and relevance to DevOps, especially in your essay proposal which seems a bit generic.
Good luck.

PR #384, opened 2020-03-19
Demo Proposal: Kubernetes self-healing

pojans:
So basically we are going to define a liveness probe for a pod; The web server running on the pod is coded in such way that we can break it with certain input and thus failing the liveness probe.


cesarsotovalero:
Your proposal looks very well focused!
Good luck.

PR #390, opened 2020-03-20
Update: Moved the presentation proposal "How to build a DevOps culture" to the right folder

johannaii:
I have mistakenly put the proposal in the old contribution folder, but I have fixed it in this PR instead.


gluckzhang:
Hi @johannaii , thanks for the update. Could you also delete the wrong folder in this PR? Thanks :)

johannaii:
@SophieHYe In my forked repo, the old folder is removed. If you see the latest commit 5b328f7, the file path got only "renamed" because I did not do any actual changes to the file. So I think the old folder in this repo will be removed too with this PR. Or what do you think? :)

gluckzhang:
@johannaii , ah I see, it could be found in the commit list. In the files changed tab it's not that obvious. Then I don't have other questions :)

PR #392, opened 2020-03-20
Add Presentation for week 2

Kappenn:
Team Members:

Gunnar Applelid
Kasper Liu


SophieHYe:
Hi,

Thanks for your PR and preliminary slides.
Automated testing indeed very related to DevOps, please also add the description in the READMe as the slides would be not accessed in the future. 

The slide in the current stage looks very general. Based on your topic, could you also detail explore the following perspectives:
- What kind of testing techniques are considered as traditional testing?
- The representative tools for these two categories?
- Scenarios of using these two type testings, their strengthens and weaknesses.
- How they benefit or impact DevOps?
- The important take away from your presentation.

Kappenn:
Our speech will complement some of these point but do you want all these point dealt with in the slides? And I've uploaded the slides as a PDF or do you mean that we write out what is said on every slide and put that in the README or just a summary?

SophieHYe:
Just a summary is fine. I merged your PR.

PR #401, opened 2020-03-23
phwe Essay proposal

phwes:


SophieHYe:
Hi,
Thanks for your PR. Your essay proposal looks great with detailed information and sources.
My major concern is about the essay scope seems a bit too general and lost focus. It would be quite hard to discuss all components of DevOps in-depth in one essay. 

To gain more original for the essay, would you be interested in add a section and focus on 'the future trend of DevOps'? or something like "DevOps for web applications /mobile applications" 
We may expect to see your insights in the essay.

phwes:
Thank you, 

I will simply choose another topic.

PR #408, opened 2020-03-24
Week 2 Presesntation: Slides

javierron:
Presentation slides link for our Docker & security presentation

### DIFF
--- a/contributions/presentation/week2/javierro-sasig/README.md
+++ b/contributions/presentation/week2/javierro-sasig/README.md
@@ -32,3 +32,5 @@ Knowing how to run secure containerized applications is an important skill given
   - Control groups
   - Docker daemon considerations
   - Linux Kernel Capabilities
+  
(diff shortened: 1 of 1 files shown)


cesarsotovalero:
@javierron  your presentation is too large. 
You can assume that everyone knows how Docker works (or have only one slide at the beginning) and focus more on the security aspects.

javierron:
Hello! The presentation will not take more than 9 minutes, most of the slides are just image ant text sequences.

 The first part should not take more than 1.5 minutes, and was added for context only.

cesarsotovalero:
@javierron you don't any security tool or real example of vulnerability in depth. I think that your presentation will improve with less introduction and more focus on real examples.

javierron:
OK, we will remove that part and add more focus on the real examples provided.

javierron:
To clarify:

Many people in the course are giving feedback saying that they don't know much about DevOps, and that's why we added some strong context.

Also that the goal of the presentation is to give an overview on the _relation of docker security and DevOps_, and not to explain any tool or vulnerability in depth.

cesarsotovalero:
Ok... good luck!

javierron:
Thank you!

PR #409, opened 2020-03-24
Presentation slides update: Git Repository Health -- As measured by the bus factor and other metrics

Mema5:
Here are our slides for today's presentation.


gluckzhang:
Got that, thanks @Mema5 !

PR #425, opened 2020-03-25
Presentation proposal, week4: Monitoring. tracing. observability in DevOps

hmellin97:
Paul Löwenström: paulher@kth.se
Henrik Mellin: hmellin@kth.se

Topic: Monitoring in DevOps

We would like to do a presentation on the monitoring phase in DevOps cycle

* what it is
* different ways of monitoring and their pros and cons.
* give an example using an existing service

### DIFF
--- a/contributions/presentation/week4/paulher-hmellin/README.md
+++ b/contributions/presentation/week4/paulher-hmellin/README.md
@@ -1,0 +1,11 @@
+Paul Löwenström: paulher@kth.se
+Henrik Mellin: hmellin@kth.se
+
+Topic: Monitoring. tracing. observability in DevOps
+
+We would like to do a presentation on the monitoring phase in DevOps cycle. 
+
+* what it is
+* what it is used for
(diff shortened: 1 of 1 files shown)


gluckzhang:
Hi @hmellin97 , thanks for the proposal. Definitely interesting and relevant. Note that there are overlaps among these three concepts, for example, monitoring and tracing could be different aspects of observability. Will you introduce the differences of them during your presentation?

hmellin97:
Hi @gluckzhang, how about if we focus mainly on the _monitor_ part and discuss its uses and available services?

gluckzhang:
Hi @hmellin97 , that sounds good. You could introduce different ways / levels of monitoring and their pros and cons instead. Could you update the `README`? Thanks :)

hmellin97:
@gluckzhang done!

gluckzhang:
@hmellin97 merged, thanks!

PR #433, opened 2020-03-26
Essay proposal: Weighting your configuration manager options: Chef and Saltstack

phwes:


gluckzhang:
Hi @phwes , thanks for the proposal, which is well written. I will merge it without questions :)

PR #442, opened 2020-03-28
Tutorial: Docker

nour95:
We didn't decide the details yet, but we both are intereseted in Docker. 
I made the pull request as draft pull request. When we are ready for the review, I will convert it to a normal pull request.


nour95:
This is the proposal for our tutorial. Is it enough to speak about the points we wrote or do we need to speak about more things? Any suggestion in this case (taking into account that we will use Katacoda)?

gluckzhang:
Hi @nour95 , thanks for the PR. Doing a tutorial about Docker is fine. However I would suggest focusing on a concrete topic, because the scope is too broad. Could you narrow down the scope in your tutorial, and give a more concrete title?

I know it might be challenging for you if you just begin to use Docker. But I still think a tutorial could contain more technical contribution, which also helps you learn more through the process.

nour95:
@gluckzhang We thought about:
- create a docker container that can listen to port 80, and show a message like "hello world" in this port.
- mounts local files to the container, So that we can see the local changes in the container.
- install a container from docker hub 
- run it 
- remove it 
- rename it 
- show the difference between a container and an image 
- show all the containers/images that had already been install in your system.

Is that good?

gluckzhang:
Hi @nour95 , I am fine with the background part. Regarding the technical part, I think it's still a bit general. The operations you mentioned could be found on many tutorials already. If your tutorial aims at beginners, I would propose to add the following discussions, in order to have more deep technical contents:

- introduce two ways of initializing a container: pulling from docker hub, building your own image (which could be published on docker hub also)
- what are the differences between `RUN` and `CMD` in a Dockerfile
- what are the differences between `bind mount` and `volume` when running a container

Note that you don't need to follow the suggestions above, the principle is to have more deep discussions instead of adding general introductions. You could still follow your original flow, but add more discussions when you mention related concepts / commands.

What do you think?

nour95:
Hi @gluckzhang 
The first point you mentioned is what we thought to do when we wrote "create a docker container (your own container) that can listen to port 80..." but it looks like what we wrote was less clear. 

Regarding the other two points, so they looks good and we would like to add them to our tutorial.
Thanks very much for your advises

gluckzhang:
Hi @nour95 , thanks for the updates. I think the proposal is now ready to be merged :)

PR #469, opened 2020-03-31
Finished tutorial: Docker & docker-compose

christinasunnegardh:
We have now finished the tutorial. Below is a copy of the README.

## Finished tutorial
**Link to the complete tutorial on Katacoda: https://www.katacoda.com/sunneg/scenarios/dockertutorial**

We have aimed to reach a "remarkable" for the 2nd, 3rd and 4th critera below.

Hope you enjoy the tutorial! :)

# Tutorial: Docker
*Team members: Christina Sunnegårdh (sunneg@kth.se), Klara Eserstam (klaraes@kth.se)*

**Interactive task:** Dockerizing a client and a server (node.js/react) in an application  separately (using docker compose).

We feel like there is no current tutorial that gives a good background on docker as well as also shows how to dockerize a client/server application and outlines when/why one would do that. We aim to provide a complete guide for this for someone who is new to Docker.

#### Structure
- Background
    - Overall info, what docker is/what it can be used for
    - Information about the task (when/why it is preferable, advantages etc, “case”)
- Tutorial part
    - Explaining commands used
    - Execute the commands on Katacoda

#### The criteria (4 necessary) we mean to fulfill are:
- The TA can successful execute all the commands of the tutorial (mandatory)
- The tutorial gives enough background
- The tutorial is easy to follow
- The tutorial contains fun facts or easter eggs
- The tutorial is doable in the browser without a local environment (needed for distinction)


gluckzhang:
Merged. Thanks @christinasunnegardh !

monperrus:
Feedback per #471

Congrats for being the very group ever doing a Katacoda tutorial for the KTH DevOps course! Overall, it's really good, all commands can be executed and they are well explaiend.

Comments:
* The first four steps are without commands, this is boring. Only the first one can be explanatory, the remaining steps must all contain commands. It could be as simple as a `cat foo.xml`, for instance to show the React configuration.
* Please explain what the example application does, who wrote it, and suggest to click on a link towards the Github repo containing it
* Step 2: there are two images, that's too many given the space on screen. Only keep one image (or merge them). In general, at most one image per step.
* Step 5: it contains too many commands, it could be split in two separate steps

Good job overall!

christinasunnegardh:
Hi,

Thank you for feedback!

1) We feel like "what is boring" can be very subjective 🤔 We would actually like to stick to the current structure: an informatory part and one hands-on part, if that is okay! We have tried the tutorial on people not familiar with Docker and they have expressively said that they liked that section. We feel like it would be too much for the reader to include commands in the informatory part, as it (according to us) should only give background on Docker - _how it works_ and not how to use it.

2) We will add a comment saying that the example application is written by the authors of the tutorial, to clarify this. The application is actually part of the surprise (Easter egg) and we therefore do not wish to tell the user exactly what it does as it would spoil our easter egg. We do however say what it does on a technical level in step 4, which we thought would be enough:
_"The client folder contains the frontend, and the server folder contains the backend of the application. When running, the server is continuously listening for a connection on port 9000. The frontend simply makes an API call to localhost:9000 and displays the response on the start page."_

3) As can be read in the text, we refer to the two images in different parts of the text. Merging these images would therefore give a negative effect on the reading experience. We therefore prefer not to do this. As the problem seem to be that it takes too much space on the screen, we will try to make them smaller - if this is an okay solution! :)

4) We thought about this, and decided that it was more neat to have the front end containerization   in one step (to avoid confusion for the reader), even if it was longer. It is of course also possible to divide it such as "Front end Part 1" and "Front end Part 2", if this is better?


With all this said, we know that we don't agree on all points - but as we do have good arguments for doing the tutorial the way we do - is it still possible to receive a "remarkable"?

We also wonder if we make a new pull request after performing the changes?

Thank you!

PR #484, opened 2020-04-01
Tutorial proposal: Dockerizing your Favorite Game

perfah:
# Dockerizing your Favorite Game

## Members
1. Per Fahlander (perfah@kth.se)
2. Joar Ekelund (joarek@kth.se)

## Proposal
Creating a Katacoda-tutorial on how to put a game in a Docker-container. The tutorial would be named “Dockerizing your Favorite Game”.

We are considering the game https://github.com/alexdantas/nsnake to demonstrate that it works.

## Motivation
Games usually have a lot of dependencies and perhaps it could be interesting to have a container that always can launch a game.

## Ties to DevOps

The project would involve the following:
- Containerization

### DIFF
--- a/contributions/tutorial/perfah-joarek/README.md
+++ b/contributions/tutorial/perfah-joarek/README.md
@@ -1,0 +1,18 @@
+# Dockerizing your Favorite Game
+
+## Members
+1. Per Fahlander (perfah@kth.se)
+2. Joar Ekelund (joarek@kth.se)
+
+## Proposal
+Creating a Katacoda-tutorial on how to put a game in a Docker-container. The tutorial would be named “Dockerizing your Favorite Game”.
+
+We are considering the game https://github.com/alexdantas/nsnake to demonstrate that it works.
+
+## Motivation
+Games usually have a lot of dependencies and perhaps it could be interesting to have a container that always can launch a game.
+
+## Ties to DevOps
+
+The project would involve the following:
+- Containerization


gluckzhang:
Hi @perfah , thanks for the PR. The idea is interesting and relevant (and quite useful especially for me :smile: ).

**A key question to be discussed:** what are the differences between dockerizing games and dockerizing other kinds of software? This is not well explained in the proposal.

Usually there are different game distributions for different operating systems like Windows and MacOS. Do you consider to dockerized a game that runs on windows, to make it running on Linux platforms? Because that is the most common scenario for dockerizing games I could think of. It's also complex and tricky (e.g., driver settings, libraries, videos/audios, etc.). In another word, I think dockerizing games which are run on Linux originally is a bit simple: we mainly use `apt-get install` to install all the libraries. I would like to accept the PR if you plan to try dockerized windows games with the help of [wine](https://www.winehq.org/) or other similar tools.

perfah:
>what are the differences between dockerizing games and dockerizing other kinds of software? This is not well explained in the proposal.
 
The original idea was that we would create a Katacoda-tutorial in which you could actually play the game at the end. It would hence have to be a terminal game (such as snake or pong) as katacoda provides no graphics that we are aware of, only a terminal instead. Terminal games are somewhat atypical in that they don’t usually share many game-specific dependencies with other games. The tutorial would however be general enough to work for most games (as it shows you how to install dependencies), but not specific enough to be significantly different from other types of software. We think this would make for a much more fun tutorial (with a reward at the end) so we wonder if your key-question is compromisable.
 
>In another word, I think dockerizing games which are run on Linux originally is a bit simple: we mainly use apt-get install to install all the libraries.
 
This is usually simple but also depends on the availability of the specific dependency library as well as the Linux distribution. Linux is very fragmented and still benefit from initiatives such as snapcraft.io (where you statically link everything) because of this. Dockerizing native games can also provide benefit in terms of making games work in multiple contexts.

>I would like to accept the PR if you plan to try dockerized windows games with the help of wine or other similar tools.
 
We are overall open to using Wine, but don’t think it would make much of a difference if we are using terminal games.

gluckzhang:
Hi @perfah , happy to hear your answer! Whether introducing Wine makes difference or not depends on the main takeaway of your tutorial. Based on your explanation, I think you could go ahead with the idea. It seems that the tutorial will focus on introducing key operations of Docker, and use dockerizing games as a concrete, interesting scenario. Correct?

When I read the title in the beginning, I think you are going to address problems like "running Windows applications on Linux using dockerization" or "dockerizing Linux GUI applications". Because this is the main purpose of dockerizing games in my life :) This also needs more introduction about X11 forwarding, audio / video configuration which may not be included if you go for a terminal game.

perfah:
Sounds great! Yes, that is exactly what we had in mind. Please tell us if there is anything else you would like us to include. Otherwise we are happy to get started! :)

gluckzhang:
> Sounds great! Yes, that is exactly what we had in mind. Please tell us if there is anything else you would like us to include. Otherwise we are happy to get started! :)

Go ahead. I will merge the PR right now :)

PR #489, opened 2020-04-02
Finished tutorial: Error Monitoring/Sentry.io

christinerosquist:
**Copy of the README:**

# Tutorial: Sentry.io

**Members of the group:** 

Christine Rosquist (rosquis@kth.se, Github: christinerosquist) and Shapour Jahanshahi (shapour@kth.se, Github: shappe)

**Topic:**
 
Cloud-based error monitoring using Sentry.io. 

We would like to learn more about monitoring and error tracking. Sentry.io seems like a beginner-friendly alternative and we want to know more about how to use it in our own applications. 

#### Tutorial:
- How to set up the platform (including installation and setup)
- How to use the platform (the dashboard etc) in your own programs 

We will provide a background with information regarding monitoring and error tracking, but also explain the main princples of Sentry.io. We will also provide a step-by-step tutorial showing the setup and how to use the platform, in a way that is easy to follow. The tutorial will include fun facts/easter eggs that gives the viewer extra information that could be fun and useful to know about. 

# The finished tutorial 
The link to the tutorial: https://www.katacoda.com/christinerosquist/scenarios/sentry-tutorial
We are aiming for the *distinction* pass. The Grading Criteria we aimed to fulfill are:

**The TA can successful execute all the commands of the tutorial:** We have tested the tutorial on different persons that all managed to successfully execute the commands, so there should not be any difference with the TA:s. 

**The tutorial gives enough background (remarkable):** We provide an introduction to Error Monitoring that do not require any previous knowledge, and we also explain why the topic is relevant in DevOps. 

**The tutorial is easy to follow (remarkable):** The tutorial is suitable for persons without any previous knowledge about Sentry/Error Monitoring, since we guide the user through each step in detail. We also provide pictures that clearly show what to do at the differnt steps. 

**The tutorial is original, no such tutorial exists on the web:** We think that our tutorial is original since we could not find any tutorial about Sentry that is doable in Katacoda. There do exist different tutorials about Sentry on the web, but they are not related to DevOps in the same way as ours. 

**The tutorial contains fun facts or easter eggs:** The tutorial contains fun facts, for instance about the history of Sentry and how it was founded. There are also many cool facts mentioned in the tutorial that can be useful for many developers (like the integration with slack and the features mentioned in the “what’s next”-part). 

**The tutorial is doable in the browser without a local environment (remarkable):** Yes! Everyone can try the tutorial in Katacoda with just the link.


bbaudry:
Feedback per #490

Congrats for a very early completion of an original task using a Katacoda tutorial! The topic is very relevant to the course, the instructions are well structured and you have set very ambitious goals on your task. The participant to this tutorial needs to switch between the Katacoda and the Sentry windows very often. I suggest adding a note at the beginning so the participant sets an easy environment for these frequent switches, e.g., open Sentry in a new window rather than a new tab. 
The tutorial is globally good.

- step 1: add link to [Sentry.io](https://sentry.io) on the first occurence
- step 2: 'add sentry package', we are not in the right folder to run cd sample-app
- step 3: there are some steps missing to reach this screen https://i.imgur.com/Ee6JVii.png
- step 4: you suggest to "make sure that you still can run the application by writing npm start"
  - indeed It failed, because it could not find this file /home/scrapbook/tutorial/package.json
  - this is because I was not in the sample-app folder
  - add instruction to go back in this folder
-  step 5 is not very interactive, which is a bit boring. 
-  step 6, 'add breadcrumb', 'app.js' is not clickable
-  step 7, I suggest an integration with Github rather than slack, since you are sure that all participants of the course have a github account, which is not necessarily true for slack (I don't have a slack account)
- step 8: is it possible to make this step interactive or let the participant perform an action / test ?

shappe:
Thank you very much for the feedback! We'll start implementing your suggestions next week. :100:

christinerosquist:
Hello! @bbaudry @gluckzhang 
We have now implemented almost all of your suggestions :) 
Since the link to the tutorial is the same, we do not need to change anything to the file in this PR.

But, regarding **step 7**, we chose to actually keep the Slack-integration part. We looked into having an integration with GitHub rather than Slack, but after some research of how the Sentry integration with Git worked - we came to the conclusion that it was a bit out of the context of this tutorial. In this tutorial we focus on finding out when something goes wrong, while the Git integration is more about traceability of what caused the error. The Slack-integration is one of the main reasons for using Sentry for Error monitoring at all - and we believe that it is such an important part that it should remain in the tutorial. However, we chose to add the Git Integration to the last step (What's Next) so that people doing the tutorial still gets aware that they also can integrate Sentry with Git. Hope you understand our choice!

Regarding **step 8**, we did not choose to make it interactive since the purpose of the step only is to provide the user of more interesting links (that is out of scope of the tutorial). The tutorial is about getting started with Sentry, so we chose to have the most essential things in Sentry as the interactive ones in the previous steps. Step 8 is more to provide more material if the user is interesting to find out even more about Sentry - after getting started with the platform.

Another note regarding the tutorial that we just want to clarify is that Katacoda does not seem to have fully support of loading files into the scenarios if the backend isn't Docker. We use Node as backend and therefore there could be problems with loading the files (not all the time tho, so you might not notice it). If you reload the tutorial, the files should show up as wanted (we also tell the user this in the tutorial - that you might need to reload the page if the files does not show up directly). But just so you know that we really tried to fix this problem, but it seems to be related to Katacoda and there is not much we could do about it. We hope that you understand this and that it do not affect the grading of the tutorial.

We are not sure how the grading is done but we hope that the grader takes the things we wrote here into consideration! I guess that @gluckzhang or @bbaudry will do the grading since you know about the feedback we got and how we motivated our changes after that. 

Best regards,
Christine & Shapour

gluckzhang:
Hi @christinerosquist , thanks for the clarification, we will take your comments into consideration when we do the grading. Happy Easter!

PR #507, opened 2020-04-06
Demo Submission: Data Corruption

stinalangstrom:
Link to the deployed [demo](https://dd2482-demo-data-corruption.firebaseapp.com/)

### DIFF
--- a/contributions/demo/tommysam-stinalan/README.md
+++ b/contributions/demo/tommysam-stinalan/README.md
@@ -8,18 +8,22 @@ Tommy Samuelsson (tommysam@kth.se) & Stina Långström (stinalan@kth.se)
 
 ### Data corruption
 
-*Data corruption is a when data becomes unusable, unreadable or in some other way inaccessible to a user or application*
+*Data corruption is when data becomes unusable, unreadable or in some other way inaccessible to a user or application*
 \- Technopedia
 
 There are many factors that can trigger data corruption, which makes it hard to protect yourself. Malware infection is one common problem but a lot of problems are due to poor software and hardware, which includes both software bugs and hardware failure.
 
 Data corruption can take several different shapes, for example file errors, including both permission changes and renaming, data incompatibility, configuration bugs and pipeline corruption (input and/or output) that cause user-facing issues. To protect against corrupted data one approach is to filter the data to prevent corrupted data from entering the system. It is also important to always have back-up on the data to be able to restore the data from a previously known good version. To protect against malware infection it is important to have good security, which includes antivirus systems but also checks of the existing data.
 
-In this demonstration we aim to show how corrupted input data, from a file, a server or a user for example, can affect a running application. We want to show the difference of filtering the input vs not filtering the input and also show a back-log where information about corrupted data is displayed. In this demonstration we mimic an event site and let a user add different kinds of input to an event and observe the result of it. This is of course applicable to almost any interactive site.
+In this demonstration we aim to show how corrupted input data, from a file, a server or a user for example, can affect a running application. We want to show the difference of filtering the input vs not filtering the input and also show a back-log where information about corrupted data is displayed, to demonstrate the consequences of an unprotected environment.
 
 ### Demo
 
-Link to the deployed [demo](https://dd2482-demo-data-corruption.firebaseapp.com/)
+Link to the deployed demo: [demo](https://dd2482-demo-data-corruption.firebaseapp.com/)
+
+Link to the screencast: [https://youtu.be/2voDWw1wz5Q](https://youtu.be/2voDWw1wz5Q)
+
+Link to GitHub repository: [Data corruption](https://github.com/Zodbigt/dd2482_demo_data_corruption)
 
 The demo consists of several options, that in general displays a new page. They are:
 
@@ -29,11 +33,11 @@ Information about this demo and data corruption in general.
 
 - **Data generator**
 
-A page which can generate input to the application. The user can choose between pre-made inputs or make an input themself. The self-made input is constructed as a filed where the inputs should be typed and radiobuttons so that the user can choose with type the input should be in (String, Int, Null or Undefined) and accroding to these options the site will behave differently. 
+A page which can generate input to the application. The user can choose between pre-made inputs or make an input themself. The self-made input is constructed as a field where the inputs should be typed and radiobuttons so that the user can choose with type the input should be in (String, Int, Null or Undefined) and accroding to these options the site will behave differently. 
 
 - **Data safe mode**
 
-The page which utilizes the data-safe-parser. Before data is displayed on this site the data goes though several check to remove data that is currupted in some way. Only data that passes the system is showed on this site.
+The page which utilizes the data-safe-parser. Before data is displayed on this site the data goes though several checks to remove data that is currupted in some way. Only data that passes the system is showed on this site.
 
 - **Regular mode**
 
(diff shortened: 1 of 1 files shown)


bbaudry:
Feedback per #508

Thank you for submitting a demo. The topic is very related to the Ops and reliability aspects of DevOps. It is great that you let the visitors interact with the demo. Unfortunately, I could not run much of the demo and the content is currently difficult to interpret and it is hard for a visitor to get clear take-aways from the demo.
- homepage opens correctly. It could be enhanced with some guidelines about you expect the visitor to do
- background page opens correctly. It can be consolidated with explicit scenario that you wish the visitor to perform. For example, how can the visitor know that the demo is happening correctly? What is the visitor supposed to see / experience on the different tabs? What are the expected learning outcomes?
- Data generation: the tab opens correctly with four possibilities to set the event name, img, date, price, description, tags and tickets. It would help to understand a bit more about these fields and the effect of choosing some values for these fields.
- safe site: sometimes I see two images about COVID and Greta, sometimes I see nothing. How do I know this is the correct behavior of the demo? How can I relate the input data generation to this behavior? 
- regular site: each time I go on this tab, the demo crashes completely. I need to close the tab and start a new tab at https://dd2482-demo-data-corruption.firebaseapp.com/
- quarantined: the tab displays lots of information. How can the visitor of your demo relate these events to the data generation? To the the safe and regular variants of the site?

stinalangstrom:
Hi!
Thank you very much for your feedback. I am troubled to hear that you could not run much of the demo. Did you experience more problems than just the "Regular site"?

The "Regular site" is supposed to represent a site without any error-handling of the input, and because of this there are situations where the entire page will crash, which is something we have made intentionally. To avoid further crashes for the Regular site, one needs to clear the database using the "CLEAR THE DATABASE" button on the homepage. We have wrote this in the readme but we do understand that it might not be that intuitive, and something we will try to improve. But hopefully if you do that, your problems with the regular site will go away.

Right now the entire site is connected to a database and everyone uses the site are connected to the same. This means that if someone else adds events that crashes the site, your site will crash as well. This is a problem we haven't really thought about before, but now when you have tried it we realise that user needs there own session instead of sharing a session. It could be the case the someone(probably us) had added corrupted data to the database before you tried it and by that your "Regular site" crashed even if you haven't done anything. This is something we of course want to solve, and we plan to implement session-cookies for that. By that a user will always have a there own set of data, making the database empty from start for first time users. So that changes they do will effect there "own" site only.

Concerning that it is difficult to interpret the content, we plan to make a screencast of the application and go through the functionality of the site. Hopefully that will help. Our main plan was to participate on the live demo session and then we could explain live if something was hard to understand, and by that we have only made the site functional, not that intuitive. (Again, something we hope can be improved by a screencast)

Our main question right now is if the content of the site is enough? The data generator and the different sites etc. We do understand that we need to work more on making a user understand what their purpose are, we tried to write that in the readme, but I understand that we need to be more clear on that. But the main content and our approach to demonstrate the problem with corrupted data, is that ok? Or do you want us to change anything?

bbaudry:
Hi @stinalangstrom 

I understand that this demo site is not standalone but meant to support your screencast. This explains a lot!
I suggest a few changes in the website to better support the screencast
- Generator : use labels that explicitly refer to parts of the safe / regular website instead of the generic 'Event' label
- in the safe site, it would help to have explicit mentions about the tactics to handle corrupted data
- quarantine: it would help to see the corrupted data value, in addition to the description of the corruption that triggered the quarantine

Good luck with the preparation of the screencast!

stinalangstrom:
Thank you! We agree with you on the changes and we will definitely take a look on them before doing the screencast.

gluckzhang:
Thanks for the updates. I'm now merging this PR.

louisezetterlund:
I will give feedback on this one :)

johannaii:
I will give feedback to this one too

felixlut:
I will give feedback here as well

GiorgosTagkoulis:
Hi, I would like to give feedback.

olapiv:
I'll give feedback here

htchan:
i will give feedback here

louisezetterlund:
Thank you for a great demo!

I love that you made an interactive part of this and that you very clearly show how to use it in the screencast. It's great that you have pre-made inputs as well :)

I also like that you have a voice-over and that you are using a QR-code. It's a nice touch!

I don't really have anything you can improve, so I'll just say: Great job!

olapiv:
Generally really high quality demo. Super easy to follow and actually quite relaxing! 

I found this a very interesting way of displaying incorrect data entries. Also the idea of restoring data entries that have been quarantined seems pretty cool. I might be naive in this sense, but I never thought about specifically checking my database log for any irregularities concerning false entries (I use a parser, no worries), but it would be interesting to see how attackers have perhaps tried to hack my sites.

 What I am missing here is the fact that you always mention this data parser - at levels have you implemented it and what are your suggestions (front-end / back-end / something else)? Also, how do you suggest saving quarantined data if it’s potentially malicious, but you still want to enable users to de-quarantine it afterwards?

htchan:
it is a great demo providing a intereactive site to try. i am curious on what is the pre-set data behind those button. It may be better if u can show what the data is when i click on the pre-set data buttons and use a unique dataset but not the pre-set data in the demo video.

johannaii:
Overall, really good screencast and thank you for sharing it with us! 

The demo/screencast is easy to follow, interactive, sublime and on a technical level that suits this course. I have one thing you could consider to change if you want to:

* It is a little bit difficult to read the subtitles sometimes. For instance, when you are choosing the pre-sets of data to use. It would be nice if you could change the position of the subtitle at those times, or use a colored but transparent background for the subtitles. Another solution could be that you also speak first, then show in the screencast which pre-set you choose.

gluckzhang:
Hi, @nicolai-h is absent today, but he also gives his feedback on this demo. Here is what we received from him.

> The demo is about data corruption, the demo shows why it is important to handle corrupted data. A regular website and a safe website is shown. The safe website uses a parser which filters out the corrupted data. The structure of the demo seemed highly thought through and it was easy to follow all the steps.  Not only is the demo easy to follow, any watcher can try it out for themselves. It would be interesting to see the code for the parser but I’m not sure if it would be important for the demo, but it would at least make it more technical. To conclude I think the demo was very good and highlighted an important topic.

GiorgosTagkoulis:
In general, I believe it is an awesome app. Thank you for sharing! 

More specifically, the screen-cast was very professional and I was able to test it in my browser while watching. Some nice features, apart from the application itself, are the QR-code at the end and the easter egg with devOps input. In my opinion, there is nothing to improve. Good job!!

felixlut:
Really good job, it's clear that you have put a lot of effort into this demo!

# Positives 
* Interesting topic, and why it's important
* The screencast was easy to follow, due to clear voiceovers + subtitles
* Shameless plug to the DevOps course (easter egg)!
* I feel like the strongest part of your demo is the interactive site! 

# Negatives
* Puts the rest of us to shame
* Improve the youtube description, I believe it's a requirement to have a short explaination of the problem + links

Zodbigt:
@louisezetterlund
Thanks for the positive feedback! Glad that you liked it!

@olapiv
The parser is on a front-end level. Generally it's good to parse the data server side as well. In this demo we wanted to show that data could potentially be corrupted on its way from a server. So by that the front-end needs to do a final check. 
The data is actually not saved, it is rather not used. The thing that is saved is just the error message/information. 
For now the site we have built does not allow a user to de-quarantine the data. But that would be a good feature for sure. 
How to handle that would be dependent on how the back-end handles the data in the first place. That depends on what way the data is malicious. If the data is wrongly formatted it is not harmful until it’s used, and by that one could have the possibility to modify it and later de-quaranteen it. But if the data is harmful, e.g. viruses, the data should not be saved and in those cases a de-quaranteen is not applicable.
Thank you for your feedback!

@htchan
Thank you for the input. We will try to figure out a way to display the example data in the video. 

@johannaii
Thanks for the positive feedback. We agree that the subtitle is hard to read at that point in the video. We will try to move it.

@nicolai-h 
Thanks for the feedback!
The code is linked in the repo for anyone to take a look if they like.

@GiorgosTagkoulis
Thank you for the positive feedback, we appreciate it!

@felixlut
Nice feedback, thank you.
We have put a lot of work into the demo especially make it interactive, glad you like it.
Thank you for reminding us about the youtube description, we totally missed that but will of course add it!

PR #515, opened 2020-04-06
ESSAY PROPOSAL: The Current State of CI/CD

GustavDowling:


GustavDowling:
This Essay will define CI and CD, and then go over what practices are common now and what tools are used and then go over some of the state of the art techniques or neglected practices that may be valuable for the future, with pro's and con's for using/not using everything.

SophieHYe:
Thanks for your PR. The proposal looks great! Maybe also compare a self-implement CI/CD tools with the on-the-shelf tools?
Your PR has been merged!

keivanm:
@SophieHYe What do you mean by self-improvement CI/CD tools?

PR #520, opened 2020-04-07
Essay Proposal: Load Balancing, and Slicer

GiorgosTagkoulis:
Load balancing is probably one of the most important parts of a distributed system. It is a pretty broad term since we might need to load-balance normal http request/responses or load-balance storage locations (including sharding). The idea is in the essay to give an introduction to load balancing in general, present the two main classes (static and dynamic) together with some algorithms in each category and at the end present Slicer; Google's general purpose sharding service.


SophieHYe:
Thanks for your PR. The essay proposal is original.
I will suggest also discuss the impact of a good design of load balancing to distributed systems from the complexity and the cost perspective and how it relevant to DevOps.
Your PR has been merged!

PR #504, opened 2020-04-06
add tutorial proposal

georgewbar:
# Jenkins pipelines using Declarative pipelines and Jenkins user interface

## Contributors
George Rezkalla - rezkalla@kth.se @georgewbar

## Description

This tutorial will be about how to use Declarative pipelines and how to use Jenkins user interface to build a pipeline to compile a demo maven java app and run the unit tests whenever the code is pushed to a repo. The idea is to show two different ways of doing the same thing.

I will try using katacoda to make this tutorial. However, if there are some issues regarding using katacoda, i will demostrate how to install jenkins locally on a computer, and do the tutorial locally.

Issue #437, opened 2020-03-26
Choosing the same topic


georgewbar:
Hi,
I have a question regarding topics. My teammate @sujon2100 and I decided to do a tutorial on Microsoft Azure. However, when doing our PR, we went through old PRs to see the template for some PRs. By chance, we found a PR that have very similar to ours. The PR #379 was about a tutorial of CI/CD pipeline in Microsoft Azure. We did not create our PR yet.

Is it allowed that two teams do tutorials on nearly the same topics or not?

gluckzhang:
Hi @georgewbar , it's fine for different teams to work on a similar topic. We believe that every team has their own insights into the topic. Plus in all cases and task categories, plagiarism is strictly forbidden, per the KTH rules which are very strict.

georgewbar:
@gluckzhang, Thanks for the answer!

gluckzhang:
The question has been answered. So I'm closing the issue now :)

PR #521, opened 2020-04-07
Tutorial Proposal: Continuous integration with Jenkins and Blue Ocean 

Volminger:
# Tutorial: Continuous integration with Jenkins and Blue Ocean 

 - Keivan Matinzadeh keivanm@kth.se @keivanm
 - Alexander Volminger alevol@kth.se @Volminger

 ## Description
We will show how to setup a pipeline using Jenkins with Blue Ocean for a Java project. We will first show how to setup and install Jenkins with Blue Ocean. We will after that show how to setup a pipeline  that is triggered each time a user commits to the connected github repository containing the Java application. We will then show how to add a testing stage to the pipeline. The pipeline will either fail or succeed if the code passes the tests. The tutorial will go through both scenarios, because the build will fail from the start so the user needs to fix the Java code to be able to make the build succeed.

We aim to do this tutorial using https://www.katacoda.com

### DIFF
--- a/contributions/tutorial/alevol-keivanm/READEME.md
+++ b/contributions/tutorial/alevol-keivanm/READEME.md
@@ -6,4 +6,4 @@
  ## Description
 We will show how to setup a pipeline using Jenkins with Blue Ocean for a Java project. We will first show how to setup and install Jenkins with Blue Ocean. We will after that show how to setup a pipeline  that is triggered each time a user commits to the connected github repository containing the Java application. We will then show how to add a testing stage to the pipeline. The pipeline will either fail or succeed if the code passes the tests. The tutorial will go through both scenarios, because the build will fail from the start so the user needs to fix the Java code to be able to make the build succeed.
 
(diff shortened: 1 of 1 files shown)


SophieHYe:
Hi,

Thanks for your PR. Your proposal is indeed the core of DevOps. FYI, your tutorial proposal is very similar to the existing one https://github.com/KTH/devops-course/pull/504. We allow you to work in a similar topic (please see issue https://github.com/KTH/devops-course/issues/437).

I will share the same concerns about this topic, which is the original and difficulty of your tutorial. Please let me know if you still want to continue this topic, if yes, please use 1-3 short sentences to state the novelty and originality of your proposal :)

Volminger:
WIP Link to essay: https://docs.google.com/document/d/1vC35eVQpjxDziZEA4dsSM52kR4jhoZIIkLpDAeLvLyk/edit?usp=sharing

We plan to publish it on Medium, but we want feedback on it first.

PR #526, opened 2020-04-08
Essay suggestion: VMs, Containers and their relationship to DevOps

jonasjohansson8908:
# VMs, Containers and their relationship to DevOps

## Members

Jonas Johansson, jonasjo5@kth.se
Github username: jonasjohansson8908

## Topic

I want to learn more about containers and how they differ from VMs. 
More specifically, I would like to cover how they can be used in a DevOps context and the
benefits and drawbacks of each solution.

## Preliminary structure

- Introduction
- Background
	- What is a VM, how does it work?
	- What is a container?
	- Images
- Differences between containers and VMs.
- Benefits and drawbacks, when to use which solution?
- A specific example covering two common tools and how they can be used in DevOps. (Will choose a specific tool of each after researching)
- Conclusions

### DIFF
--- a/contributions/essay/jonasjo5/README.md
+++ b/contributions/essay/jonasjo5/README.md
@@ -1,0 +1,24 @@
+# VMs, Containers and their relation to DevOps
+
+## Members
+
+Jonas Johansson, jonasjo5@kth.se
+Github username: jonasjohansson8908
+
+## Topic
+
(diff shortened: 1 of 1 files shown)


SophieHYe:
Thanks for your PR. Your essay proposal is great! This PR has been merged!

PR #532, opened 2020-04-08
Demo: Adding monitoring to an API for a personal website

jadlers:
I will add monitoring of stats like response time and status for requests sent
to my API. The monitoring will be visualised with something like grafana or
prometheus. This will allow me to create a better understanding of how the API
is used. I will also add persistent logs in order to make it easier to trace
errors.

The demo will show how the data monitored is captured and also how it's
displayed.

### DIFF
--- a/contributions/demo/jadlers/README.md
+++ b/contributions/demo/jadlers/README.md
@@ -1,0 +1,20 @@
+# Demo
+
+## Members
+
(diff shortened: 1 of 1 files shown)


gluckzhang:
Hi @jadlers , thanks for the proposal. Monitoring and visualization is very important in DevOps. The idea is interesting and relevant. The PR has been merged.

PR #539, opened 2020-04-10
Tutorial proposal: Apache on docker

phwes:


gluckzhang:
Hi @phwes , thanks for the proposal. I'm afraid that only talking about deploying Apache server on Docker does not contain sufficient technical contribution. Maybe you could consider to add the following ideas:

- also introduce network communications among containers, because a web application usually connects to a database server, which could be dockerized as well
- introduce the differences between `volumes` and `bind mount`

What do you think?

phwes:
@gluckzhang That sounds great!

I will do that as well then.

gluckzhang:
@phwes , perfect. Please adjust the `README` according to our discussion. Then I will merge the PR :)

phwes:
I have now updated the proposal.

gluckzhang:
Thanks @phwes . The PR has been merged.

phwes:
Great.
Thank you for your help!

PR #555, opened 2020-04-13
Finished tutorial on chaos in Kubernetes

jvikstrom:


gluckzhang:
Thanks for the submission. Merged!

monperrus:
Thanks for the original tutorial and congrats for going the extramile on katacoda. It's the first time that I see kubernetes in katacoda and chaos engineering in katacoda as well, that's really cool!

Major comments:
- make all commands copy-pastable automatically (see https://www.katacoda.com/docs/scenarios/markdown-syntax for example the commands in https://www.katacoda.com/sunneg/scenarios/dockertutorial)
- make the katacoda linear
- explain the content of config files upfront, before using them

More precisely:
- step 1: explain what is deployment.xml and what it contains. explain what the pods are doing.
- "The other way": what is "the initial way"?
- Step 2: "service-accounts.yml" tell more about what it contains
- "And thedeployment" typo, use a spellchecker and grammar checker everywhere
- "a message that kube-monkey has terminated a nice-pod" be precise: which message exactly?
"And this is an example of a config file." which one? add "You can see it with "cat foo.yml".
- comment the YML files as well
- "labels using vim t": avoid using vim, many people don't know how to use it and would be lost. One option is to instead propose a one-line sed command that changes an existing configuration file
- " this has also been preconfigured in single-kill-deployment.yml" avoid. A katacoda is clearer with it is linear, choose one single path to avoid confusion (either sed or single-kill-deployment.yml)
- Step 6: "this command below" explains this complex command.

PR #564, opened 2020-04-15
Demo Proposal: Using Github Actions to run tests on Docker containers.

olapiv:


olapiv:
Okay, for some reason it shows that I changed 75 files.. I'll just do it again with a new clone.

PR #565, opened 2020-04-15
Demo Proposal: Using Github Actions to run tests on Docker containers

olapiv:
We have one question about this - I created a docker-compose project for my tutorial with another team-partner (see [here](https://github.com/olapiv/docker-nginx-tutorial)). Would it be possible to "fork" my own repository and for this demo? Since this demo isn't really about how to use Docker/docker-compose in general, but more about seeing how GitHub Actions works with Docker.


olapiv:
Actually, we just realized that we will not have to create a public repository for the demo.. since we have a live project, which is perhaps a bit cooler, we would then use that. It does however, in general, have a fairly similar setup to the repository for the tutorial that I mentioned. We would be fine doing a video on that, but not make all the code public.

We have meanwhile also decided that we would use Postman's CLI client Newman to do API-testing on the Docker-setup within GitHub Actions. The tests would be written within Postman. If the tests go through, we would then push the built images to their corresponding Dockerhub registry.

cesarsotovalero:
@olapiv: Your demo proposal is very technical and strongly related to several DevOps topics. 

> since we have a live project, which is perhaps a bit cooler, we would then use that

Ok, perfect.

> It does, however, in general, have a fairly similar setup to the repository for the tutorial that I mentioned.

That's not a problem at all.
 
> We would be fine doing a video on that, but not make all the code public.

Can you prepare a sample of your project that you could make public? The purpose of the demo is that everyone could reproduce it. So it's very important to create something that others can reproduce.

> We have meanwhile also decided that we would use Postman's CLI client Newman to do API-testing on the Docker-setup within GitHub Actions.

Postman is a good tool for the task.

SophieHYe:
Your PR now is good to be merged!

PR #566, opened 2020-04-16
Finished Essay: The correlation between SaaS and DevOps

christinasunnegardh:


monperrus:
Thanks for the first version.

Comments:

- there is no clear takeaway
     - tell the reader what's the most interesting parts of your essay in a bullet point list at the end of the introduction
- avoid orphan subsection
- an essay must have a top-level conclusion
- Discussion section: identify three concrete points to be discussed, put them as subsection, ground the discussion on the case study

See also the comments in the annotated PDF sent over email.

PR #571, opened 2020-04-16
Update COVID-19 project proposal

balwes:
My group is near a final submission so the project has changed slightly and some things have become more clear. The proposal now reflects these things. Most importantly the last feature has been replaced.

### DIFF
--- a/contributions/covid19/bwest-rug/README.md
+++ b/contributions/covid19/bwest-rug/README.md
@@ -17,6 +17,6 @@ Educate the user on the COVID-19 spread using a SIR model.
 * Make predictions from data based on mathematical models such as SIR-model
 * Allow the user to view the data in any date interval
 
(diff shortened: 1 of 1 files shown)


gluckzhang:
Hi @balwes , before merging this PR. Could you elaborate a bit on why you replaced the last feature? Is it because of technical problems or something else?

balwes:
Sure. Basically, we thought it fell out of scope due to time constraints. We found it difficult to estimate this kind of data and time could be better spent on other features.

Instead, we are adding a customizable date interval because having to see all the data at once isn't very readable. This feature increases user-friendliness and therefore becomes helpful to more people.

gluckzhang:
@balwes thanks for the explanation. It makes sense to adjust it then. The PR has been merged.

PR #574, opened 2020-04-17
Canary Releases

sashikanthr:
Changed our presentation topic to be more narrow. We will be discussing only Canary Releases as part of Continuous Experimentation.


SophieHYe:
Thanks for your update. The PR has merged!

PR #579, opened 2020-04-18
Fix directory names

xmas92:
We / I used my Github username instead of our/my KTH email prefix. This PR updates the directory names to the correct values


cesarsotovalero:
Thanks for the fix @xmas92

PR #592, opened 2020-04-19
Demo Proposal: Load balancing gRPC Remote Procedure Calls in Kubernetes

felixlut:
We would like to demo the topic specified in the proposal!


gluckzhang:
Hi @felixlut , thanks for the proposal, which sounds very interesting and relevant. The PR has been merged!

PR #593, opened 2020-04-19
Updated Tutorial Link: Implementing CICD in ServiceNow using flow designer

akhilyerrapragadaa:


gluckzhang:
Thanks for the contribution!

bbaudry:
Feedback for #593

Thank you for the early contribution. The tutorial is well structured but can be better motivated and can be more engaging for the reader. Please see the comments below to consolidate your submission.

- Introduction paragraph
  - move the short summary and motivation for ServiceNow early
  - what is 'spoke'
  - why is important to have the ability to move your application?
  - what are the competitors?
- Creating an application 
  - what is an application here?
  - what is the purpose of an application on ServiceNow?
  - "Assume that the Developer instance we are using is the Test instance" -> this means that there can be different types of instances. What are the other types? What are they used for?
- You explain how to perform different types of tasks. Why these tasks? Why are they specifically interesting? What should the reader get out of these tasks
- Add a conclusion for the tutorial

Typos
- ServiceNow which with -> ServiceNow  with

PR #598, opened 2020-04-19
Tutorial completed: Fuzzing with Radamsa

Mema5:
Hello,

Here is the final version of our tutorial on fuzzing with radamsa.


bbaudry:
Feedback for #598

Thank you for this early contribution. The tutorial is relevant, technical, well-structured and comprehensive. I appreciate the use of katacoda, the Java examples are good, the additional resources are useful. Overall this is a solid contribution to the Devops course. Below are some comments to make the tutorial completely self-contained.

Step 1
- you suggest 'echo "hello" | radamsa' to check that install is fine. It would help to also indicate the expected behavior for this command so if someone runs it, she knows if the behavior of the command actually indicates that install is fine. I had to ctrl-c, not sure if this is correct.

Step 3
- sometimes the bad input looks correct, e.g., a regular integer such as 52. Is this expected?
- Typo: the script has the message "Crash found! Bad input:", but the message on the console is "Exception found! Bad input:"

Mema5:
Hi @bbaudry, thanks for your feedback!

Step 1: you're right. I explained the expected behaviour. 
Step 3: actually you found interesting bugs and that's exactly what we want to achieve thanks to Radamsa! I made it more clear by changing a little bit the script and explaining more.

The changes related to this feedback are visible in [this commit](https://github.com/Mema5/katacoda-scenarios/commit/c2fb1ea8ea12ec70b01941be3e2f4a347ffa11a9).

Mema5:
We are done with this contribution on our side. We could even receive some feedback from Aki Helin, the maintainer of Radamsa! I think that the PR can be merged. Thank you!

gluckzhang:
> We are done with this contribution on our side. We could even receive some feedback from Aki Helin, the maintainer of Radamsa!

Excellent!

> I think that the PR can be merged. Thank you!

Merged. Thanks for the contribution :)

PR #607, opened 2020-04-20
Essay Proposal: A/B Testing and its benefits

johansettlin:
We would like to learn more about A/B testing and its benefits.


cesarsotovalero:
@johansettlin Good topic for an essay. Please also cover its drawbacks and make sure to reflect your own opinion about A/B testing.

PR #611, opened 2020-04-20
Essay proposal: Software Measurement and Code Coverage

richarduggelberg:


SophieHYe:
Thanks for your proposal. In the essay, we always expect to see the linkage to the academic such as the state-of-the-art in code coverage, and their perspective limitations, etc. I would also look forward to reading some literature review of this topic in your essay.

Your PR has been merged!

PR #629, opened 2020-04-21
Presentation Final sumbission: Git Hooks: What They Are and How to Use Them

Goradux:
Added the slides in PDF format, as well as the outline.


gluckzhang:
Thanks for the updates, merged.

PR #631, opened 2020-04-21
Final essay submission: An Analysis of AWS with Respect to DevOps Characteristics

Goradux:
Have changed all of the suggested parts of the essay according to the feedback. Have also fixed the folder name as well, since we forgot to include our 3rd member there.


SophieHYe:
Thanks for the update!

PR #634, opened 2020-04-21
Add link to presentation proposal

emetha:
We have added the presentation link to our README


SophieHYe:
Thanks for the update!

PR #627, opened 2020-04-21
Tutorial with martin

htchan:
i seperate the delete and insert of folders into two commit, does it delete the onld one (tutorial/htchan) now?

PR #638, opened 2020-04-21
delete duplicated folder (htchan)

htchan:
i change the htchan to htchan-mgabri, so here is the pull request to delete old one
requested by SophieHYe in https://github.com/KTH/devops-course/pull/627


SophieHYe:
Thanks for the update！

PR #646, opened 2020-04-22
Demo proposal

MindaugasVarkalys:
# Demo proposal

## Students
Mindaugas Varkalys varkalys@kth.se @MindaugasVarkalys  
Justas Dautaras dautaras@kth.se @sonderangebot10

## Topic
Deploying apps on Kubernetes cluster using Rancher pipelines

## Motivation
Rancher is a great tool which simplifies management of Kubernetes cluster through the intuitive UI.
It also has pipelines feature, which allows to run CI/CD jobs and see their status on the same UI environment. Having everything at the same place is a great benefit, so our demo will show the setup of Rancher pipelines as well as explain the main features of Rancher UI. As an example, we will demonstrate the continues deployment of a simple api to Kubernetes cluster using Rancher pipelines.

### DIFF
--- a/contributions/demo/varkalys-dautaras/README.md
+++ b/contributions/demo/varkalys-dautaras/README.md
@@ -1,0 +1,14 @@
+# Demo
+
+## Students
+Mindaugas Varkalys varkalys@kth.se @MindaugasVarkalys  
+Justas Dautaras dautaras@kth.se @sonderangebot10
+
+## Topic
+Deploying apps on Kubernetes cluster using Rancher pipelines
+
+## Motivation
+Rancher is a great tool which simplifies management of Kubernetes cluster through the intuitive UI.
+It also has pipelines feature, which allows to run CI/CD jobs and see their status on the same UI environment.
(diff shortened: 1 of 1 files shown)


cesarsotovalero:
@MindaugasVarkalys: Cool topic for a demo. Do you have plans to use Katacoda or any other teaching platform?

MindaugasVarkalys:
@cesarsotovalero Katacoda or similar tools are great for tutorials, but in my personal opinion, it's not the best way to demonstrate a working demo, especially when it includes the UI. So we are planning to make a Youtube video. In the first part of the video we will quickly explain the main features of Rancher and show how to setup the pipelines. In the second part we will show a working demo of continues deployment. Of course, the first part can used as a tutorial, but the main reason of including it in the screencast is to make our demo repeatable by even the ones not familiar with Rancher. Since some setup steps and our demo includes the UI, the best they to demonstrate that would be a video, isn't it?

cesarsotovalero:
@MindaugasVarkalys Thanks for your answer. A video backed up with supplementary material is the best approach for your demo.

PR #649, opened 2020-04-22
Tutorial Proposal: Version Control

pojans:
##### Participants

| Name | Email |
| ------ | ------ |
| Pojan Shahrivar | pojans@kth.se |
| Richard Uggelberg | rug@kth.se |

##### Proposal
We want to create an online tutorial which takes the user through the process of contributing to an open source project.

The aim is to expose the user to all of the git commands you need to know in a realistic scenario that tie them all together.

##### Target audience
This tutorial is aimed at students that have experience with coding, but not necessarily has used version control to participate in development of a large project.

### DIFF
--- a/contributions/tutorial/pojans-rug/README.md
+++ b/contributions/tutorial/pojans-rug/README.md
@@ -1,0 +1,16 @@
+#### Tutorial Proposal: Version Control in Large-scale projects
+
+##### Participants
+
+| Name | Email |
+| ------ | ------ |
+| Pojan Shahrivar | pojans@kth.se |
+| Richard Uggelberg | rug@kth.se |
+
+##### Proposal
+We want to create an online tutorial which takes the user through the process of contributing to an open source project.
+
+The aim is to expose the user to all of the git commands you need to know in a realistic scenario that tie them all together.
+
+##### Target audience
+This tutorial is aimed at students that have experience with coding, but not necessarily has used version control to participate in development of a large project.


cesarsotovalero:
@pojans: Thanks for the proposal. Notice that there are plenty of tutorials and posts about the importance and the first steps for contributing to open source (eg. [[1]](https://www.yegor256.com/2015/12/22/why-dont-you-contribute-to-open-source.html) or [[2]](https://github.com/firstcontributions/first-contributions)). So, if you want to go in this direction, you need to be more precise and narrow down the focus of your tutorial. For example, how to address a specific OSS community? or a specific set of OSS projects? or how to deal with a particular contributing scenario?

If your focus is more on Git techniques for this, then I'm afraid that it's very hard to come up with something original there. In this case, I suggest you to change the topic.

pojans:
@cesarsotovalero We are aware that topic is not original in any way, but we are hoping that the scenario that we have in mind and the way we present the solutions to the problems that arise will be original and I think the tutorial will be _very useful_ in the end.

The scenario deals with something I can confidently say I am not alone in doing wrong, and I've had to look up several tutorials to solve the issue. Also I don't to give up too much in this proposal as there is a humerus part to it as well that I don't want to ruin.

In the worst case, it is going to be in an unoriginal tutorial, but you will be able to follow and execute all commands, it's going to be easy to follow and without a local environment.

cesarsotovalero:
@pojans OK, but you should be aware that passing the originality grading criterion may be difficult, according to the topic of your tutorial.

What kind of environment or teaching tool (eg. Katacoda) do you plan to use?

pojans:
@cesarsotovalero Absolutely, thanks for notifying us about that! :) 

We plan to use Katacoda or a similar online env that allows us to guide the user through each of the commands step by step.

cesarsotovalero:
@pojans Good luck with your tutorial.

PR #652, opened 2020-04-23
Essay proposal

ashgreyship:
My essay proposal was approved last month but it was deleted yesterday.


SophieHYe:
Thanks for the PR, your proposal is good to merge!

PR #666, opened 2020-04-24
Tutorial: Change proposal- Remove bind mount from tutorial proposal

phwes:
I have found that the current difference between bind mount and volumes are small and confusing. For example, dockers own documentation both recommends using and not using bind mounts. Most probably because they are currently being changed. 

I was wondering if it would be okay if I removed the comparison between volumes and bind mounts from my tutorial?


phwes:
[This](https://www.katacoda.com/phwes/scenarios/apache-docker) is the state of the tutorial at the moment.

SophieHYe:
Thanks for the update! Nice progress!

PR #668, opened 2020-04-24
Essay Proposal: Microservice Architecture

felixlut:
# Essay Proposal
## Member
- Felix Luthman (felixlut@kth.se)

## Topic - Microservices
With the increased usage and complexity of web applications, i.e. applications that aren't installed on the user's machine, but rather deployed on a remote server, the microservice architecture has emerged as a solution to a lot of the problems a Monolithic web application may face. I would like to explore how these two models are structured, how they differ, and when one should pick one over the other.

## Initial structure
* An introduction to what the Microservice architecture is, how it works, etc
* How it differs from the Monolithic approach
* A comparison of pros and cons between the two
    - When one should be favored over the other
* Why this matters from a devops point of view

### DIFF
--- a/contributions/essay/felixlut/proposal.md
+++ b/contributions/essay/felixlut/proposal.md
@@ -1,0 +1,13 @@
+# Essay Proposal
+## Member
+- Felix Luthman (felixlut@kth.se)
+
+## Topic - Microservices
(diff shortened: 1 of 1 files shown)


SophieHYe:
Merged!

PR #685, opened 2020-04-26
Tutorial Proposal: Docker for Machine Learning using Nvidia Docker

luic-c:
This tutorial aims to explore how to  run machine learning by utilizing Nvidia Docker, which includes steps to install Nvidia Docker, commands to run Nvidia Docker


SophieHYe:
Thanks for the proposal. Indeed, a GPU docker benefits deep learning.
When developing the tutorial, it would be good to always think about the originality of your work and the differences between the existing GPU docker tutorial.

PR #690, opened 2020-04-27
Finished Essay: VMs, Containers and their relation to DevOps

jonasjohansson8908:
# VMs, Containers and their relation to DevOps

## Members

Jonas Johansson, jonasjo5@kth.se
Github username: jonasjohansson8908

## Topic

I want to learn more about containers and how they differ from VMs. 
I would also like to cover how they can be used in a DevOps context and the
benefits and drawbacks of each solution.

## Structure

- Introduction
- Background
	- Virtual Machines
	- Containers
		- Container Orchestration
- Differences between containers and VMs.
	- Storage
	- Performance
	- Security
	- Portability
- VM, container, or both?
	- Combining the tools, a more specific example
- Conclusions


SophieHYe:
Thanks for the update! Merged.

PR #695, opened 2020-04-27
Requesting Feedback for Tutorial

nagavem:
Completed the tutorial on Testing Microservices in Spring Boot Applications: Tools and Techniques and requesting feedback for the same.


SophieHYe:
Hi @nagavem,

It is too late to arrange feedback for today(April 27).  Could you please move your feedback planning date a few days after? Thanks!

PR #728, opened 2020-04-28
Finished demo: ci/cd pipeline with k8s and travis-ci

nicolai-h:


jvikstrom:
I'll give feedback here

Marcus9512:
I will give feedback here : )

Pontus-Broberg:
I would like to give feedback to this!

olapiv:
I'll give feedback here

Marcus9512:
Hello!

Here is my feedback to your demo.

First of all I would like to say that it is good that you slowly went through the files, this gave the viewer a chance to read and understand what they do.  The screencast also looks nice!

However, the first question that appeared in my head is what is this video is about, what tools are you using and why are you doing this? The tutorial starts right with a demonstration, but I think it would have been nice if you introduce CD and what tools you are using. This is also a grading requirement, “The demonstration is clearly motivated (why it matters for Devops?)”

I like that you take time to show all the steps after the push, but it get´s slow when we are waiting for the tests to pass and the page to update, maybe speed up this part?

It maybe would be nice if you explain what your new patch did, because I did not understand what the difference was. I liked the easter egg : D

The most important comment I have is that you did not have a take-home message, which is required by task in order to pass (see the grading criteria) or at least I could not find it.

In your github repo, maybe add a readme so that I know how to use it?

Sometimes the subtitles are a bit hard to read, maybe add a black border around them?

Then again, nice demo, this is just some suggestions on how you could improve it according to me.

Regards Marcus

jvikstrom:
First off: Music's great I love it.

Second: I think you should consider doing something like [this](https://youtu.be/dInVpRP6de8) to have a background to the subtitles, just makes everything look a bit more clean.

Considering this is a CI demo, at around 0:48 you use https git instead of the ssh pub/private key, I really think you should redo it with the ssh key auth instead, it'll just look 10x more professional.

Also, you don't really explain the travis yaml, consider to spend some time in the begining and walk through the most important points of it.

Also, your Youtube video is missing a description. I think that was a requirement, so you should strongly consider looking into that. Another thing is "take-home message", I think that is a requirement, but I can't seem to find anything in the criteria so I might be wrong about that.


I think you have a really nice pacing in the video, I feel like I have no problem reading everything and figuring out what you're doing. Which again brings me back to the music, it really helps with that feeling as it's just such an incredibly chill song.

Visually this looks pretty good, nothing is cluttered at least. If you want to improve your chances of having the "visually impressive" criteria I'd suggest you style the index.html page to look better (I have no idea how to do design though)

The ending is absolutely amazing: I love how you just suddenly end at that.

Pontus-Broberg:
Hi! Here's my feedback.

First of some kind of introduction to the subject would be nice to see. You can explain what we are about to see, what Travis and K8s are, and why you are doing this. 

White text on a white background is hard to read. It is good you have outlined the letters with a black border but maybe consider adding an entire background box with a different color. It would make the reading of the subtitles much easier. Also, you could consider zooming in on the code and console a bit to make it easier to read as well. 

The entire section between 1:10 and 3:30 can be fast-forwarded in my opinion since basically nothing happens during this time. It is not very fun to stare at the Travis outputs. If you do not want to fast forward this, at least you should consider adding some kind of text while it runs. You can talk about Travis, what it does, what it is good for, and so on. Or you could give the viewer some kind of fun facts etc. But as of right now, the majority of the video is really boring. 

In the other parts of the demo, the pacing was good and it was clear what you were doing. So keep this clarity and pacing while taking some of the feedback to account and you will have a great demo. Good Job!

olapiv:
I understood from this demo that you can somehow use Travis to automatically deploy things.

What I’m missing however:
* What and where is k8?
* How is Travis set up / connected to Github?
* Maybe something spoken as well? I find it hard to read the subtitles and watch your cursor go through the files at the same time.

What I personally really see as unnecessary:
* The whole Github commit, login, push thing
* Filming the build process

dieflo4711:
@olapiv @Pontus-Broberg @jvikstrom @Marcus9512  Thank you for your feedback!

PR #739, opened 2020-04-29
Final submission: COVID-19 Assignment

johannaii:
Fixed the feedback comments and added the images for the report.


Zodbigt:
I just want to highlight that the file COVID19_COURSE_INFO.md does not display two images correctly.
But the same markdown code shows the images in our repository:
https://github.com/Zodbigt/dd2482_covid19_course_info/blob/master/COVID19_COURSE_INFO.md

It's the **firebase_nodes.png** and **test_playground.png**.
You can check that they are submitted in the code and in the associated “images” folder.

SophieHYe:
@Zodbigt 
Hi, this is a minor issue. We can accept a PR to fix them. Thanks!

PR #765, opened 2020-04-30
Demo Proposal: Using Terraform with GCP server for static IP webhook calls

akhilyerrapragadaa:


gluckzhang:
Thanks for the proposal. Merged.

PR #779, opened 2020-04-30
Essay draft

eywoon:
Draft for the essay.
I have a lot more for the body (around 1500 words) but think gives a clear idea of how I think about the essay so some feedback on structure would be really appreciated :)


eywoon:
Why do I have these old commits? it's a separate branch merged from upstream master, do I need some hard reset?
I want them away 😅

SophieHYe:
Hi, @eywoon 
This PR is fine. If you don't want the old commits in the PR, you can use git cherry-pick to only submit specific commits in the PR.

eywoon:
Yeah of course @SophieHYe 
I never use it, just squash, good time to start :)

PR #787, opened 2020-05-02
Update: New link to screencast

Callet91:
Added names on title screen and "Thanks for watching" in screencast. Also some minor fixes to make the text more readable during section regarding Hardware tests.


SophieHYe:
Hi @Callet91,
Thanks for the update. Could you please take a look and fix the conflicting file? Thanks!

jadlers:
I'd like to give feedback to this demo :relaxed:

nour95:
Hi, Can I give you feedback?

cwinge:
Hi, I'll give some feedback for this demo :)

stinalangstrom:
I'd like to give feedback to this demo as well!

cwinge:
Feedback time!

Overall it was a very nice screencast. Everything was easy to see and follow, and some nice music in the background (missing credits, unless I somehow missed it?).

The highlighting of the key parts of the screen at times, was a nice touch. 

Easy to read the subtitles (color contrast) and good duration to be able to read everything without pausing.

Quite nice to have actual footage of the hardware used included.

Not sure if you included an easter egg or not, it you did I missed it. Unless it was the finger in the last footage, but not sure what it references.

Only real improvements I have would be the following (only minor things, some not required):

Might have wanted to have the bookmarks toolbar hidden, and close unused tabs, as well as remove the downloaded file at the bottom.

There are some weird screen movement at some parts, unless it is an added effect? But looks like the sceen is moved slightly at some points, showing other parts of another monitor? Happens slightly at 0:56ish and can see part of a terminal later on, to the right.

2:23 "Skiping" -> "Skipping"

2:32 different tenses. "build" -> "built"

-----
That was all I had for now :) gj

jadlers:
Over all the video is well made and I think it meets all mandatory requirements, good job! :+1: 

Some comments on good parts and possible improvements:
- Adding an initial overview could make the demo easier to follow.
- Nice highlighting of areas to help the viewer know what to look at.
- When showing the hardware device [here](https://youtu.be/_xO6UTt_2I0?t=95) (1m35s) I don't quite follow if the software test run on PlatformIO's designated software test servers are running the tests on such hardware?  It confuses me as the next step is the hardware step where you state that Jenkins uses a raspberry pi as a slave load to which it uploads the appilcation.
- It's nice that you show every step and don't skip any parts of the pipeline. Makes it easy to follow how it should look when the pipeline is working properly.
- The use of music is nice since there is no recorded voice. The song also fits well as background music to a demo like this. Credit to the artist?
- The explanatory box always contain a good amount of text. However, a possible to improvement could be to add voice over. I had to pause in and rewind a few times in order to be able to read the information and also see the screencast. This in not a problem since it's easily done, just a pointer to a possible enhancement.

stinalangstrom:
An interesting demo, good job! Here comes my feedback:

* The subtitle on the right side is visually appealing but, in my opinion, it is hard to read at the same time as you watch the screen. For me it would be better if the subtitle is more like "normal" subtitle (in the bottom of the screen), since it is easies to both read and watch at the same time then. Or you can "pause" the screen for a few seconds so that one can read the information before you start, and then keep your subtitle as it is (it looks great, so I suggest you to do that).
* I liked that you darkened the screen to make it clear where to look, very good.
* Very nice that you introduced hardware in this demo, good job! It made it very interesting.
* Sometimes it was a bit hard to see what was going on in the programs. You could add small notes that explains it, for example "running all the tests..." etc, just to get the user to know what is happening when the program runs.
* I liked that this really was a demo, it showed what happened but did not explain too much about how. Good job!

nour95:
Hi, 
I think the video looks very good and it very clear and sublime. 
I just think that since there were no voice (and also due to the position of the subtitles), I was forced to stop the video to read the the sublime. Maybe it is needed (the time for each sentence show is enough as @cwinge mentioned but I was stressful and wanted to read the text and see what is written by the tools in the video at the same time). It will be really better if you add a voice to the video that says what is written in the subtitles.

nagavem:
I would like to provide some feedback for this demo

nagavem:
Feedback:
Found your demo to be very interesting especially as I had never considered the possibility of using DevOps Principles with respect to embedded devices,it made me click on your demo instantly on reading the same.The visuals and presentation are nicely done and spot on.

In terms of improvements just a few:
as @stinalangstrom and @nour95 mentioned, I too found reading the subtitles and focussing on the demo at the same time to be difficult. 
The music chosen is nice,however the volume could be decreased a little at times as it is taking away attention from the content.
While the build activites are taking place there are gaps in the video where nothing else is occurring so this can be either filled with subtitles or trimmed.

Overall a well explained and very interesting demo on a topic that I will definitely explore more into thanks to this.

xmas92:
@cwinge @jadlers @stinalangstrom @nour95 and @nagavem 
First of thanks for all the feedback and glad that you got some enjoyment from our demo.
One common theme here is to try and give the viewer more time (to both read and absorb what is happening) and make it feel less stressed. We will try and rectify this as much as possible. Full credit to @Callet91 that is editing this. 

Also there were a few questions about the music. It was originally not credited at the request of the creator, but you've convinced him to leave anonymity. So we introduce this original song created by none other that @Callet91 himself.

@jadlers 
> * When showing the hardware device [here](https://youtu.be/_xO6UTt_2I0?t=95) (1m35s) I don't quite follow if the software test run on PlatformIO's designated software test servers are running the tests on such hardware?  It confuses me as the next step is the hardware step where you state that Jenkins uses a raspberry pi as a slave load to which it uploads the application.

There are two test phases here. The first test compiles a test suite (set of unit tests) that is uploaded to each bare device. They then communicate the results (essentially result of every assertion) over the serial USB interface. This is used to test internal logic on each device.

The second test is about testing the actual application. Here the normal application is uploaded to the device under test (DUT). We then have the other Arduino board which has a program that can take commands over USB. The Raspberry PI then sends commands to this board telling it to send some voltage on the DUTs input pin (simulating a real sensor value), and then commands the board to print the voltage values that it observers on the DUTs output pins. (The LED brightness values). 
The breadboard in the middle is a simple opamp circuit to convert PWM duty cycle to voltage values, part of the "testing rig". This test the external behaviour of the device.

We used the extra Arduino board in the middle as it was more complicated to create something that interfaces the Raspberry PIs 3.3v pins with the Arduinos 5v pins. 

In a more serious test rig the Arduino board would be replaced by test equipment like oscilloscopes, source meters, power supplies and communicated with and controlled using some standard like [LXI](https://en.wikipedia.org/wiki/LAN_eXtensions_for_Instrumentation) or [GPIB](https://en.wikipedia.org/wiki/IEEE-488).

We discussed just this issue before making the demo on how to explain the test setup so it made sense  and fit within the 5 min window. We will however add a more detailed explanation (including information on how to recreate the demo) in the GitHub repo readme.

PR #791, opened 2020-05-03
Added demo URLs

cwinge:
Updated with demo URLs

Screencast URL with easter egg hint: https://youtu.be/VHttX3IZOhw (might be hard to spot due to speed up)

Screencast URL with easter egg shown: https://youtu.be/XTmD-rRdCK4

Github repo: https://github.com/cwinge/web-terminal

URL to the demo will be posted on the 5th as a comment on the pull request. The easter egg can be used with the demo.

### DIFF
--- a/contributions/demo/cwing/README.md
+++ b/contributions/demo/cwing/README.md
@@ -1,9 +1,9 @@
 
 ## Updated with demo URLs
 
-Screencast URL with easter egg hint: https://youtu.be/H8R7EtumyL8
+Screencast URL with easter egg hint: https://youtu.be/VHttX3IZOhw
 
-Screencast URL with easter egg shown: https://youtu.be/Xd505ZyVmzw
+Screencast URL with easter egg shown: https://youtu.be/XTmD-rRdCK4
 
 Github repo: https://github.com/cwinge/web-terminal
 
--- a/contributions/demo/cwing/README.md
+++ b/contributions/demo/cwing/README.md
@@ -1,7 +1,9 @@
 
 ## Updated with demo URLs
 
-Screencast URL: https://youtu.be/vfoIpUxWGOQ
+Screencast URL with easter egg hint: https://youtu.be/H8R7EtumyL8
+
+Screencast URL with easter egg shown: https://youtu.be/Xd505ZyVmzw
 
 Github repo: https://github.com/cwinge/web-terminal
 
--- a/contributions/demo/cwing/README.md
+++ b/contributions/demo/cwing/README.md
@@ -1,3 +1,12 @@
+
+## Updated with demo URLs
+
+Screencast URL: https://youtu.be/vfoIpUxWGOQ
+
+Github repo: https://github.com/cwinge/web-terminal
+
+----
+
 ## Inspiration behind the demo
 Just as a short background, I figured I'd give the reason I thought of doing this demo.
 
@@ -41,10 +50,3 @@ Could also prepare customized images for demos/presentations needed to be perfor
 
 ### Concerns
 My main concern is how well the free tier services available might be quite detrimental to how responsive the terminals will be, especially with multiple users, but I would not think it is the fault of the idea. Could alternatively publish the website on my own desktop instead of AWS/Azure(will still be how it is done in the actual demo) during the 5th, but will see if it will be needed or not.
-
-----
-## Updated with demo URLs
-
-Screencast URL: https://youtu.be/vfoIpUxWGOQ
-
-Github repo: https://github.com/cwinge/web-terminal
--- a/contributions/demo/cwing/README.md
+++ b/contributions/demo/cwing/README.md
@@ -41,3 +41,10 @@ Could also prepare customized images for demos/presentations needed to be perfor
 
 ### Concerns
 My main concern is how well the free tier services available might be quite detrimental to how responsive the terminals will be, especially with multiple users, but I would not think it is the fault of the idea. Could alternatively publish the website on my own desktop instead of AWS/Azure(will still be how it is done in the actual demo) during the 5th, but will see if it will be needed or not.
+
+----
+## Updated with demo URLs
+
+Screencast URL: https://youtu.be/vfoIpUxWGOQ
+
+Github repo: https://github.com/cwinge/web-terminal


cwinge:
The URL to the interactive demo is: http://35.228.222.20:8080/

felixlut:
I can give feedback here

AristotelisKotsias:
Hi, I would like to give feedback to this demo.

xmas92:
I can give feedback.

stinalangstrom:
I can give feedback to this as well!

Paul-Philip:
I'd like to give feedback here!

AristotelisKotsias:
Thank you for your demo. It is nicely done, explanatory and with a constructive take-home message! :)

Something to keep in mind is that instead of showing the coding process you can instead focus on the important parts of the code and explain in more details, through subtitles, why you do it like that (maybe) and why it matters. In general, your subtitles could contain more information.

Paul-Philip:
Thanks for the Demo, I think the topic and result are really interesting, well done setting it up!

I believe the pace was a bit fast, it's also me reading a bit slowly...
The whole coding in hyper-speed and then setting things up in the cloud while adding text strips explaining what was happening was a bit overwhelming.
Since you're showing exactly what to do to set this up step by step it's almost more like a really fast tutorial?
Perhaps you could focus on the most important parts while setting it up.
Explaining which key components are created and why they are needed for this to work and then showing them live what they do.

All in all I liked the result a lot and it's a really cool tool!
Nice easteregg! :)

monperrus:
@AristotelisKotsias could you elaborate? could you list three concrete actions to be taken to improve the demo? (you can update your initial feedback)

cwinge:
@AristotelisKotsias Thank you for your feedback. 

The code part is one of the parts that I generally feel the most insecure about adding at all. I felt like it would be nice to include it briefly, if someone would want to check any of the code without visiting the github repo. But as you said, some parts are obviously not that interesting to view(which is why I speed it up so much). Not sure if any parts really are all the interesting to point out and focus on, but I'll definitely consider it. Thanks!

@Paul-Philip  Thank you as well for you feedback.

The pace at certain parts are something I also feel are a bit too fast at the moment. 
As I stated above to Aristotel, I did the coding hyper-speed mostly to be able to quickly show what code is actually needed to make this work (or more so, the parts, python, html, bash etc). I might decide to remove the setup and code parts, and have more time used for the VM setup etc, allowing me to slow it down a bit, and maybe add a bit more information regarding those steps. 
Yeah the tutorial point is definitely true, omitting the setup would change that, good point. 

Thanks for the valuable feedback, the last couple of points included. :)

stinalangstrom:
Thank you for a nice demo, it was very interesting to watch it! Here comes my feedback:

* It was a good motivation in the beginning, one wants to know more about this. 
* The whole demo was very visually appealing, well done!
* The text are short and explains well what is happening, good job. Some of the colors could on  the other hand be changed, for me it is a bit hard to read yellow on top of blue for example. But the red and white was good. You could keep the same colors throughout the whole demo if you want too, I did not really understand why they colors changed.
* I liked the part when we got to see you write code. Of course it was hard to see what was written, but the effect was nice and you could understand more what it takes to actually do the setup. And since you post the Github repository one can easily look that up if interested, very nice. 
* You wrapped things up nicely in the end, you should probably add a more concise take-home message to make sure you fulfil the grading criteria.

cwinge:
@stinalangstrom  Thank you for you feedback.

Nice to hear someone liked the code part!

The reasoning behind the color change, was to have the background of the subtitles be vastly different to the main color of what is happening in the background. For example I use red background with white text for the latter half of the demo, but the blue one for the former due to the terminal being dark red. Very good point about the yellow text on blue. For me on my normal screen, I can read it super well and it pops out from the blue easily, but now when I checked on another monitor it was a lot harder. Will definitely change this. Thanks!

I'll consider adding something more to the end, thanks for the valuable feedback. :)

felixlut:
Thank you for this demo, I found it very interesting!

## Positives 
* I really liked your choice of background music 
* Regarding the showing of the codewriting process, I think it's fine. Since it's so sped up, it's clear to me that I'm not supposed to try to replicate it straight up, but instead read your short comment explaining what the file is for 
* Anything with SpongeBob gets a pass from me
* Good quiality overall

## Negatives
* You could add voice to the demo
* Right now you have 4 take-home messages, you might want to norrow it down a bit

xmas92:
# Mandatory Criteria Check

- [X] The screencast is 3-5 minutes
- [X] The screencast contains subtitles which are clear and in proper English
    - Very good. It is generally hard to to say just enough to make the actions on screen clear, you really pulled that off.
- [X] The screencast contains a good and concise take-home message at the end
    * Wrapping up page. Good conclusion of the project.
    * Maybe some more general take-home message can be made about how it is possible to create your own version of tools out there. Or something along those lines. Or a sales pitch for the product you created 😄. 

* The demonstration screencast is clearly motivated (why it matters for Devops?) 
    * I think this is done first by motivating the creating and then showing its overview and how it fits in the DevOps toolchain. Maybe the graders wants you to be even more explicit when explaining how it fits in the overall DevOps, I'm unsure. 
* The demonstration screencast is difficult to do
   * Definitely get the feel like a lot of work went into this demo. 
* The demonstration screencast is original 
   * I could not find anything quite like this. 
* The demonstration screencast is sublime (eg visually appealing) 
   * The text and mouse movements are choreographed well and helps with seeing what is going on. You already use a large cursor, however one thing that could make it even more clear is to add some highlighting or framing for when we as a viewer should take notice of something. Like when navigating the web interface or reading some crucial data like the IP.
    * Some of the text the font could be larger, for the code segment it does not matter to much as it is mostly there to be visually appealing not to show you how to code, but the web terminals all are a little to small to be a comfortable reading experience on my laptop screen.
* The demonstration screencast contains an easter egg
    * The easter egg was cute. I guess we are suppose to explore it ourselves? I think you could have that docker container just run and flash by in the end when you are demoing without adding extra text that says `Easter Egg`. 
* An accompanying Github repository has been made (optional)
   * Very nice. Shows how to set the demo up as well. Could benefit from a link to the screencast to show it in action.

PR #808, opened 2020-05-05
Demo: Add youtube link

pojans:
## Screencast
The screencast has been made available on [youtube](https://youtu.be/AO7T9LJGoGc). Enable the CC option to get subtitles.

##### Easter Egg
There is an easter egg in the screencast that should be visible to the perceptive person, but in case you've missed it. Pause the video around 2 minutes and 20 secs and take a look in the left side.

##### Easter egg 2
To "Unlock the secrets of longevity", please visit [http://35.228.240.95:3000/](http://35.228.240.95:3000/) and have a go.

### DIFF
--- a/contributions/demo/pojans-akshays/README.md
+++ b/contributions/demo/pojans-akshays/README.md
@@ -2,8 +2,7 @@ Demo Proposal
 ## Members
 
 Pojan Shahruvar (pojans@kth.se)
-Avijit Roy (avijit@kth.se)
-Akhshay Sinha (akshays@kth.se)
+Akshay Sinha (akshays@kth.se)
 
 ## Topic
 We want to make a demo of the self-healing feature within Kubernetes. We are designing and implementing a web service which the viewers can interact with make requests to; and some of these requests will tank the web service, which triggers the self-healing. 
@@ -12,3 +11,12 @@ We have some original ideas on how to trigger the self-healing and are currently
 
 ## Tech
 Currently, we're looking into using Google Kubernetes Engine within the Google Cloud. Web service will be coded in Node.js.
+
+## Screencast
+The screencast has been made available on [youtube](https://youtu.be/AO7T9LJGoGc). Enable the CC option to get subtitles.
+
+##### Easter Egg
+There is an easter egg in the screencast that should be visible to the perceptive person, but in case you've missed it. Pause the video around 2 minutes and 20 secs and take a look in the left side.
+
+##### Easter egg 2
+To "Unlock the secrets of longevity", please visit [http://35.228.240.95:3000/](http://35.228.240.95:3000/) and have a go.


AristotelisKotsias:
Hi, I would like to give feedback to this demo.

Anthial:
I'll provide some feedback for this demo!

juliuscc:
I will provide feedback

gynther-k:
I will give feedback on this, will be the fourth one, but it seems interesting...

AristotelisKotsias:
An interesting topic that it is well explained and backed-up with a good example! The technical parts are well covered and important techinical information are pointed out. I don't have any important imporvement to add other than that in the iterative example you could you the same number (99) as in the recursive example.
Good job!

juliuscc:
The demo is great! I like clear cut examples like this. You chose a specific but interesting topic and showed it in an interesting way. The web page looks great, as well.

I think that sometimes there was a disconnect between the content on the screen and what you said. For example, on the second to last slide, you mentioned three benefits of K8, but you didn't list them on the screen. It would have been a clearer take-home message if you used both visuals and sound to convey your message.

Slides 2 and 3 were a bit confusing if you only look at them. They made sense when listening to you but, because of the confusing wording, they didn't improve anything. For the second slide, you could have written:

```
- Hard to manage moving parts

When a small part breaks you have to:
- Detect the error
- Route around the breaking part
- ...
```

This way it is much clearer what you are talking about. Otherwise, I think the slides are very nice visually and I love the gif on slide 3.

At about 2:17 you show the yaml-file for K8 and discuss it. However, you only mention one of the keywords on-screen. It would have been valuable for me if you mentioned more concretely what you had configured. If you need more time there is a break at 11:10 where you can get a couple of seconds.

I think the demo was remarkable on many points, and especially if you fix the minor points I mentioned.

gynther-k:
Feedback:

Very nice demo overall and easy to follow. It is short and consise and nice accent and voice to listen to. Possible improvements could be that the audio quality is acceptable but could be even more improved with a high quality microphone. In 1:08 where the author says "Let's take a closer look in at the eeehrm c c loud console" it seems like he forgot what to say. Just record that sentence again for making it smooth. 2:45 "The documentation is pretty clear about different options and what
they do" Is it possible to talk a little bit more about these other options?

Anthial:
## Feedback
I believe a bit more video / audio editing could be done to make this good screencast shine even brighter! There’s a good amount of background noise and there are some long pauses for example between 01:09 and 01:16 where activity essentially stops. Now I’m not saying that these pauses are wrong, sometimes pauses are really useful for allowing viewers to treat the information they’ve been presented with, but if you want an intentional pause reducing background noise would be lovely. 

At 01:23 you go into the liveliness demo then quickly go deeper into it. You might want to explain what you are doing at that time, or simply skip the page in-between. I got quite confused the first time it happened and had to rewatch it just in case. 

Other than that I think that you have provided a great demonstration of Liveness, its uses, and the use of health checks for any distributed system. Not only that but the usage is well motivated in the introduction and in the ending, giving the whole screencast a feeling of being complete. I would have loved to see the take-home message on screen though, rather than just the question presented during the message. 

The demonstration is clean and visually well-structured, with text that when shown is treating the same area of what is being said (other than the subtitles which obviously match almost perfectly) allowing consumption of the information from two angles. Not only that but there is a good GIF in the beginning of the presentation, explaining the problem being demonstrated in the most simple terms possible (which was really useful for the introduction to this area). 

Really good job! Also, that Easter Egg! It gave me a good laugh after I couldn’t find it by skimming through the video. Well played.

pojans:
@anthial @gynther-k @juliuscc @AristotelisKotsias

Thank you for the comprehensive feedback. 😊
We will go back to the drawing board and try to solve the issues that were highlighted.

Thanks! 😀

dieflo4711:
Nice demo! It was easy to follow and contained examples. It was also visually appealing in my opinion, and technical. The audio quality could be improved (it’s acceptable nonetheless).

PR #813, opened 2020-05-05
Changed demo folder name from Github usernames to KTH usernames

olapiv:


Anthial:
I would like to submit feedback for this demo :)

louisezetterlund:
Me too :)

johannaii:
I would like to give feedback to this one too :)

jzafar:
I would like to give feedback!

Paul-Philip:
I'll submit feedback here! :)

louisezetterlund:
Hi,

Thank you for your demo!

Here are some suggestions for improvement:

- I also think you should add subtitles, it's a bit hard to follow. 
- Everything moves pretty fast. Maybe you could show the code when you're actually changing something in the code otherwise, you could show presentation slides with bullet points?
- In the end, it would be nice with a conclusion.

johannaii:
Good demo, and thank you for sharing it with us! 

I think the subject you have chosen is really interesting, and overall I think you have made a great job. As you say in the end of video, it is a lot you need to go through. However, I think you have covered the most of it in the demo. I have some suggestions for improvement: 

* The voice over is really good. However, you have no subtitles. Even if you can use the autogenerated subtitles in Youtube, it would be better if you added your own subtitles. 
* You should add a take-home message in the end.

jzafar:
Nice demo showing a great workflow of postman. Speed/Pase is good. 
i would like to add couple of things
- Add subtitles with voice.
- Add Presentation slide at the beginning  of what you are going to give demo
- Add Presentation slide at the end with

matsskoglund:
Clearly motivated since it shows how to use newman to test an API in a CI/CD pipeline
Screencast was good, perhaps as you  said yourself, a lot of material to cover and thus perhaps a little difficult to follow. But good overall.

Small comments
The postman collection file is hardcoded inside the newman docker image that was created. Perhaps it would be easier to 
use the public postman/newman docker image and then pass the collection as a parameter instad. Now the docker image must be rebuilt
each time the tests change.

I like the fact that someone is doing a demo of testing, because testing otherwise always ends up having low priority.

Good job!

Paul-Philip:
Hi, thanks for the demo!

I belive the pace is great and it is easy to follow, eventhough there was quite some information to take in.

It looks like it's is a great tool for creating testing for APIs right?
Is it possible to make any performance testing based on the generated newman files?
Also, all the responses were 200 OK in the result, is it possible to write tests for 4XX and 5XX codes which pass as well?
Perhaps just mentioning some other possible use-cases or limitations of the tool would be great!
Good job!

Anthial:
## Feedback

There is a good amount of content being covered in this demonstration and things happen very quickly. In that sense I believe an introductory diagram could be helpful in easing viewers into the very detailed setup that is being presented. 

Other than that there is no introductory slide nor ending slide (with say a take-home message), both which I think could be useful for your demo. 

Although the YouTube auto-generated subtitles are quite alright, they do miss things and a modified version of those would be really appreciated for clarity. 

Other than that the presentation is well structured and clear, both in what is being shown on screen and what is being said, going through each step thoroughly, something which at least gave me an understanding of what was being done. It is also really nice that you have a repository if one wants to essentially try this out. 

If you had an easter egg, it must have been well hidden because I went through the demo twice, skimmed through it a third time and couldn’t find it.

Great job overall!

olapiv:
First of all, thanks for all the feedback guys!! :) I'll just start right off answering some questions..

@louisezetterlund, @johannaii, @jzafar, @Anthial 
**Subtitles**: Absolutely. We were just a bit scared of having to entirely re-record the video and because it's so much content with so much text, we decided to rely on the auto-generated subtitles for now (which were surprisingly good imo). We'll fix it though.

olapiv:
@louisezetterlund, @johannaii, @jzafar, @Anthial 
**Intro & Outro**: Noted. We'll try to come with something :)

olapiv:
@Anthial 
**Easter Egg**: Sorry! We didn't think of one yet :)

olapiv:
@matsskoglund
**Hard-coded Postman Collection**: Really good point. What we did not explain in this demo, was that we also tested some APIs for uploading images and audio files (to AWS S3). These files have to be copied into the Docker image, so Newman can access them. Therefore it was necessary for us to create a custom image in any case. The workaround we have for this in the live project, is that we use Github Actions' caching functionality. There we cache the entire build process and so build time is literally halved. We posted some info to that in the corresponding Github repository.

Technically you could also just use a public URL of your Postman collection, which is super neat. However, we wanted the request collection json files to be part of the version control.

olapiv:
@Paul-Philip
Since the video is already super dense, we would find it difficult to add more information. Here are some answers to your questions though.

**Is Postman good?**: Generally - yes. I really like this way of testing since I know my backend will hold what it promises, also if it can be a bit difficult to debug. However, what I found was a major downside, was that every collection sort of has it's defined run-cycle (First do request "Register", then "Login", then "Upload Image", etc.) and that you can currently not re-use any API-request for multiple collections. So I have to copy "Register" and "Login" into every possible workflow when using my backend, which essentially means building quite a few collections with duplicate requests. We personally managed to stick to a single collection, which handles all workflows, but it is definitely not optimal.

**Checking status codes**:
Other status codes - absolutely. What is run currently in the test is:
```
pm.test("Status code is 200", function () {
    pm.response.to.have.status(200);
});
```
So just adjust the codes. And yes, all tests are written in Javascript within the Postman app.

They have a list of pre-defined tests, which you can super easily just add.

**Performance testing**: We didn't look into that intensely. As far as I know however, Postman is not meant for concurrent load testing. You can however, write tests that will check whether a response has taken longer than x ms. You can also run multiple rounds for your collection. But all requests are sent one after another.

PR #828, opened 2020-05-06
Final demo screencast submission

cwinge:
Updated screencast based on feedback.


SophieHYe:
Merged. Previous feedback could be found here: https://github.com/KTH/devops-course/pull/791

PR #833, opened 2020-05-07
Test Amplifcation: Addressed Feedback

sashikanthr:
Addressed the following feedback points:

- Broken down the subtitles as they have occupied more screen space.
- A dedicated Take Home Message.
- Minor points like credits, adding about DSpot in video description


SophieHYe:
Merged.

PR #856, opened 2020-05-11
Update folder name

emetha:
Accidently wrote the wrong folder name for open source contribution, created this PR to update it.

Folder name "etham-phwe" in open-source folder should change to "emetha-phwe".


gluckzhang:
Hi @emetha , I think you should change the folder names of the other three tasks. Note that a folder is name with team members' **kth email account**.

PR #860, opened 2020-05-12
fix: Minor bug in our COVID19 report

johannaii:
We had a minor bug in our report, where the path for some of the images were wrong. We have mention this in a comment in our previous PR #739. Got an OK to fix this from the TA:s.


SophieHYe:
Merged and thanks!

PR #868, opened 2020-05-18
Name fix

nagavem:
Spacing issue in the name caused it to be displayed as clickable link,fixing the same.


gluckzhang:
Merged. Thanks @nagavem for fixing this!

PR #877, opened 2020-05-20
Repeat Tutorial Complete

rbratfors:
# Tutorial Complete Repeat: TeamCity

## Group members
- Robin Bråtfors, bratfors@kth.se
- Kasper Liu, kasperli@kth.se

## Description
A tutorial that teaches the reader how to install and setup TeamCity from scratch for a gradle project on GitHub. The tutorial will have some screen captures and should be easy to understand for anyone with experience in GitHub, but not necessarily with continuous integration.

## Submission
The tutorial is posted on:

https://github.com/Kappenn/TeamCity_Tutorial

We have a fun fact in the background section and also an easter egg if one chooses to use the provided repo for the tutorial.

## Repeat:

We have removed the tutorial from Medium and now it exist only on github.

We have expanded the background to explain about how CI and automated build tools are connected to Devops and why TeamCity is a good option.

We have added instructions to install Java 8 and Gradle, either from the respective website or through a package manager.

### DIFF
--- a/contributions/tutorial/bratfors-kasperli/README.md
+++ b/contributions/tutorial/bratfors-kasperli/README.md
@@ -1,4 +1,4 @@
-# Tutorial Complete: TeamCity
+# Complete Tutorial: TeamCity
 
 ## Group members
 - Robin Bråtfors, bratfors@kth.se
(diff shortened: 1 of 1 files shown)


SophieHYe:
Thanks! your updated tutorial is merged.

PR #885, opened 2020-05-29
doc: acknowledge FIP

bbaudry:
added a short description of fip, with a few references. what do you think?


monperrus:
Great, thanks @bbaudry and thanks @radiofrance's FIP

PR #894, opened 2020-05-31
Updated essay for repetition

AristotelisKotsias:
The essay has been updated according to the feedback. The changes that have been made are:

- [x] Introduction now focuses on Maven instead of the benefits of build tools
- [x] Updated figures
- [x] Comparison between Maven and Ant has been removed
- [x] Bibliography now includes more scientific papers and books
- [x] Personal reflection on the pros and cons of using Maven has been added


monperrus:

PR #899, opened 2020-05-31
Updateate README.md

pojans:
#### Change log

**Title:** Changed title to better reflect on the content of the essay.
**Intro:** Reduced length and focused more on learning outcomes and content of essay.
**References:** Adopted an in-line referencing style with links.
**Methods:** Added penetration testing, linked to all methods.
**Methods & Tools:** Added discussion about which methods and tools and the  criteria for selection.
**Conclusion:** Rewritten to reflect the on all points that have been brought up in the essay & tie the knot together.

**Comments:** We decided to keep the references in the end, even though, we prefer to remove them to better fit in medium. We did this because we want to be 100% sure that the references criteria has been met. We have now used two styles of referencing which should satisfactory. We feel like the article fulfills the basic requirements to be passed and we hope you feel the same.

### DIFF
--- a/contributions/essay/pojans-sraa/README.md
+++ b/contributions/essay/pojans-sraa/README.md
@@ -16,4 +16,15 @@ We want to learn more about continuous testing and therefore we chose to write a
 - Conclusion
 
 #### Finished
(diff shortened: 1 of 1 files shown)


monperrus:

PR #901, opened 2020-05-31
Final Essay Submission

hittesharma:


bbaudry:
ACK. received repetition for essay and tutorial.
WARNING: it is risky to submit the repetitions for 2 different tasks with a single PR

hittesharma:
I could not realise that this pull request is also having tutorial submission alongwith Essay.
My intention was only to submit Essay here. So kindly ignore the tutorial (because I have already passed Tutorial) and this is just for Essay submission.

PR #928, opened 2021-03-23
Course automation: Dashboard for tasks

Ackuq:
# Course automation: Dashboard for tasks

## Members

Axel Pettersson (axp@kth.se)
GitHub: [Ackuq](https://github.com/Ackuq)

Christopher Gustafson (chrigu@kth.se)
Github: [ChristopherGustafson](https://github.com/ChristopherGustafson)

## Proposal

Create a GitHub action that is triggered when new tasks / proposals are committed to the repository, this will then generate a dashboard of all the contributions made overall, including those submitted in the previous years.

The requirements for the actions is:

-   Automatic trigger for new submissions
-   Include the authors and title of the task
-   Link to the task
-   Classify if completed or not
-   Group by year and category
-   Display the number of submissions per category / year

### DIFF
--- a/contributions/course-automation/axp-chrigu/README.md
+++ b/contributions/course-automation/axp-chrigu/README.md
@@ -10,7 +10,7 @@ Github: [ChristopherGustafson](https://github.com/ChristopherGustafson)
 
 ## Proposal
 
-Create a GitHub action that is triggered when new tasks / proposals are commited to the repository, this will then generate a dashboard of all the contributions made overall, including thoese in the previous years.
+Create a GitHub action that is triggered when new tasks / proposals are committed to the repository, this will then generate a dashboard of all the contributions made overall, including those submitted in the previous years.
 
 The requirements for the actions is:
 
--- a/contributions/course-automation/axp-chrigu/README.md
+++ b/contributions/course-automation/axp-chrigu/README.md
@@ -1,0 +1,20 @@
+# Course automation: Summary markdown for each category
+
+## Members
+
+Axel Pettersson (axp@kth.se)
+GitHub: [Ackuq](https://github.com/Ackuq)
+
+Christopher Gustafson (chrigu@kth.se)
+Github: [ChristopherGustafson](https://github.com/ChristopherGustafson)
+
+## Proposal
+
+Create a GitHub action that is triggered on commits to categories which traverse through the submissions from inside the each category and generate a markdown file which summarizes all of the submitted tasks / proposals.
+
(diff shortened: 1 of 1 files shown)


cesarsotovalero:
Hi @Ackuq.
An action like you're proposing is already merged, see https://github.com/KTH/devops-course/blob/2021/contributions/course-automation/felixfon/README.md

Can you please modify your proposal?

Ackuq:
Hi @cesarsotovalero,
When we made our proposal we were the first ones to propose this, wouldn't it be first come first served?

cesarsotovalero:
Hi @Ackuq, 

There are various TAs handling PRs at the same time. 

> wouldn't it be first come first served?

It is first merged first served.

ChristopherGustafson:
Hi @cesarsotovalero ,

I really don't think this is fair. During the lecture professor Martin stated on multiple occasions that it was first come first served. The matter of having multiple groups wanting the same automation was discussed during the lecture at which point we realised that we had to be quick and created a pr on this topic before anyone else since Martin said that it would be first come first served. "First merged first served" was never mentioned. 

Maybe you can talk to Martin about this as I am very sure that this is what he stated during the lecture.

cesarsotovalero:
Hi @ChristopherGustafson, 

I understand your point. But it is simply very difficult for us as TAs to handle and checking all PRs topics at the same time. 

Can you please consider slightly changing your action?

For example, instead of a summary per type of contribution to produce a dashboard comparing the contributions made at different years? This will be both original and useful.

ChristopherGustafson:
Hi @cesarsotovalero 

I guess we have no choice at this point than to change, so yes we can change to your proposed automation. Follow up question on this, what type of information would be useful to have in the comparison of the different years?

I also want to make a suggestion that you go over the PRs in the order they were submitted, that way no later PR would be accepted before an earlier. I don't think this is a very hard problem to solve. And when Martin so clearly stated first come first served during the lecture it is annoying to have to change ideas when we have already started thinking about a solution and even had to shift our focus for a second from the lecture to have time to write the pull request.

Ackuq:
@cesarsotovalero I've updated the PR with the adjusted scope of the task.

cesarsotovalero:
Hi @ChristopherGustafson, 

> I also want to make a suggestion that you go over the PRs in the order they were submitted, that way no later PR would be accepted before an earlier.

Ack. Thanks for the remark.

cesarsotovalero:
LGTM. I'll merge.

PR #951, opened 2021-03-23
Demo: Bitbucket clone.io integration

Sebberh:


cesarsotovalero:
hi @Sebberh, your demo proposal is promising. I'll merge your PR.
I suggest you look at previous years' demos to have a clear idea of what we expect.

PR #958, opened 2021-03-23
Course automation: Notification for proposal-PR new comments

isacarvid:
# Course automation: Email notification for proposal-PR new comments

## Members
Ayub Atif (aatif@kth.se)
GitHub: [ayubatif](https://github.com/ayubatif)

Isac Arvidsson (isacarv@kth.se)
Github: [isacarvid](https://github.com/isacarvid)

## Proposal

We want to send a notification via email to all PR group members when a new comment on a proposal-PR has been made(new comment, approved, request changes). The problem arises as only the author of the PR would receive a notification normally, leaving the other members without knowledge of needed changes. Thereby providing an improvement to the course automation experience.

The requirements for the actions are:

-   When a proposal-PR is commented it triggers Github actions.
-   Parse README.md to find all members.
-   Include PR-status and title of PR.
-   Include the new comment.
-   Send email to all group members.

### DIFF
--- a/contributions/course-automation/isacarvid-aatif/README.md
+++ b/contributions/course-automation/isacarvid-aatif/README.md
@@ -10,12 +10,12 @@ Github: [isacarvid](https://github.com/isacarvid)
 
 ## Proposal
 
-We want to send a notification via email to all PR group members when a new comment on a proposal-PR has been made(new comment, approved, request changes). The problem arises as only the author of the PR would receive a notification normally, leaving the other members without knowledge of needed changes. Thereby providing an improvment to the course automation experience.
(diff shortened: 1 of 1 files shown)


isacarvid:
An alternative approach would be to instead provide notification of a proposal-PR review's result (approve, request changes). This assumes that the review functionality is used for proposals 😁

cesarsotovalero (approved):

cesarsotovalero:
Hi @isacarvid, the originally proposed functionality is good. I suggest you give the students the choice to use the action or not. For example, you can allow using some wildcard word for setting on/off the action (which should be off by default). This way, students who don't want to be notified are not going to receive email alerts.

cesarsotovalero (approved):

PR #993, opened 2021-03-25
Presentation Kanban vs Scrum

evkade:


Deee92:
Hi @evkade, looking forward to your presentation! A discussion about concrete examples of organisations where these methodologies are applied, and how they are useful, would be very cool.

evkade:
Hi @Deee92 ! We were thinking about making the presentation in week 5 instead of week 4. Is it ok to change to do that? Then I will just make a new pull request changing the location of the folder with our readme.

Deee92:
Hey @evkade, since week 5 is pretty open right now, go ahead!

evkade:
Thank you! Making the new pull request now

PR #1003, opened 2021-03-26
Essay: Comparison of Azure, AWS, and GCP based on DevOps.

MDRezaulHasan:


khesoem:
Hi @MDRezaulHasan ,

Thanks for the proposal. Your PR changes more than one files. Please submit a PR that changes only one. Also, your chosen topic for the essay is relevant and appropriate. However, I think you should be a bit more specific about the metrics you want to use for the comparison. I'm looking forward to your enhanced version of this proposal....

MDRezaulHasan:
Hi Khashayar,
"I think you should be a bit more specific about the metrics you want to
use for the comparison", actually we are planning to compare our essay
based on CI/CD pipeline and cloud computing.
So should I mention these things inside our readme file and send another PR?

khesoem:
> So should I mention these things inside our readme file and send another PR?

Yes, please. The more detailed you are, the better it is. For example, you can say "we will compare them based on how easy it is to set up a .... CI in them. we also compare their limitations" and ....

PR #1004, opened 2021-03-26
Presentation: The history of devops

hallon-heyman:


SophieHYe:
Thanks for your proposal.  Your presentation title "The history of devops"  seems too board for your content which is about the Git.  Could you please update your title or content accordingly?  Maybe also discuss the tools integrated into Git, such travisCI, etc.  We always welcome concrete examples given in the presentation.  Please update your PR accordingly. Thanks!

hallon-heyman:
@SophieHYe Content isn't about the git anymore, check the updated readme:

"Theh history of DevOps

Our intention is to to give our viewers a deeper understanding of why and how devops as a concept, practice and profession came to be in the last decade.

 Slide draft
 - The problem of separating ops from dev
 - the growth of devops as a concept
 - Important influencers and their contribution
 - the creation of devops as an academic and professional subject and role
 - the future of devops"
 
 We swapped the presentation from git as apparently that was better suited for an essay.

SophieHYe:
Thanks for the update. The new proposal looks great. I am merging it now.

PR #1008, opened 2021-03-26
Course automation: Essay bibliography check & summary

amarhod:
### Members
Amar Hodzic (amarh@kth.se)

Natan Teferi Asegehegn (ntas@kth.se)
### Proposal
We would like to create a Github Action that produces a PR comment with the number of references used as well as an excerpt from the reference list. This would simplify the TAs job in getting an overview without having to open the file and scroll down.
We are aware of another proposal that summarizes essays but it has no mention of references.

What the Action should do:
* When a PR uses a specific label (e.g. essay), create a comment on that PR
* The comment should have a reference count
* The comment should include the reference list
* The reference list should be stripped down based on verbosity level set in the code (e.g. remove "accessed date", URL etc)

Criterias it should fulfill:
* Done before April 6
* The task produces a PR comment
* The task runs on a standard platform (Github Action)
* The task is praised by the other students of this course
* The code for the task is available and well documented

### DIFF
--- a/contributions/course-automation/amarh-ntas/README.md
+++ b/contributions/course-automation/amarh-ntas/README.md
@@ -1,0 +1,21 @@
+# Course automation: Essay bibliography check & summary
+## Members
+Amar Hodzic (amarh@kth.se)
+
+Natan Teferi Assegehegn (ntas@kth.se)
+## Proposal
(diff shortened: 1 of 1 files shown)


SophieHYe:
Hi,
The summary of essay bibliography looks interesting. That will a nice reading list for all the students who love to explore the DevOps as much as possible. If would be great you could compared with the previous years essay as well, then we can see how many new/state-of-the-art work are reference in this year's essay.

 I am merging your PR and look forward to your contributions. Thanks.

PR #1018, opened 2021-03-26
Essay proposal: SaaS, PaaS, IaaS? A taxonomy of cloud service models

andreaskth:
# Essay: SaaS, PaaS, IaaS? A taxonomy of cloud service models

## Members

Andreas Kärrby (karrby@kth.se)  
Github: [andreaskth](https://github.com/andreaskth)

Eleonora Borzi (borzi@kth.se)  
Github: [EleonoraBorzis](https://github.com/EleonoraBorzis)

## Proposal/motivation

For new people who are just learning about the cloud, it can be daunting when most learning material often targets professionals who already know about the "why" of the cloud, and just need to know about the "what" and "where" (i.e. "this tool can help you do X, and you could leverage Y for the Z-part"). The vernacular is oftentimes dense with terms thrown around that may just confuse a beginner even more ("___ as a service" is a popular one), and the problem with learning about it is "where to start"? Not knowing which pieces of the cloud-puzzle goes where could for example make the beginner try to understand why one would use Docker and when to use Kubernetes, while in reality these things cannot be compared in this way (in the 'dev' of 'devops', you could liken it to comparing e.g. BFS and DFS to Python, which of course wouldn't make sense since two are graph search algorithms, and the other is a programming language).

Therefore we would like to write an essay that we would have liked to read when learning about these things. To narrow the scope we want to focus on different cloud service models, how they work, how they are similar and different, their use cases (from the perspective of a budding business, for example), and maybe some other things if we think of them.

These are topics we plan to explore (more might be added if it will fit the word criteria of 2000 words)

- Introduction (some history; the move from on-premise deployment to cloud deployment; why; advantages; drawbacks; general comparison)
- Description of the most common cloud service models (SaaS, etc. (as above in title) and how they differ from each other; popular tools available)
- If there's space left, one could describe 2-3 concrete tools more in detail and relate them to the "bigger picture" of cloud services
- Conclusions/discussion/forecast for future

## Sources
https://www.lexology.com/library/detail.aspx?g=6aaba427-d980-4938-856a-4a1308444cc2  
https://www.bigcommerce.com/blog/saas-vs-paas-vs-iaas/#examples-of-saas-paas-and-iaas  
https://bestarion.com/iaas-vs-paas-vs-saas-enter-the-ecommerce-vernacular/

### DIFF
--- a/contributions/essay/borzi-karrby/README.md
+++ b/contributions/essay/borzi-karrby/README.md
@@ -1,0 +1,27 @@
+# Essay: SaaS, PaaS, IaaS? A taxonomy of cloud service models
+
+## Members
+
+Andreas Kärrby (karrby@kth.se)  
+Github: [andreaskth](https://github.com/andreaskth)
+
+Eleonora Borzi (borzi@kth.se)  
+Github: [EleonoraBorzis](https://github.com/EleonoraBorzis)
+
+## Proposal/motivation
+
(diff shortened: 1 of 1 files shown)


Deee92:
Looking forward to reading your essay, good luck!

PR #1020, opened 2021-03-26
Essay Proposal: Comparing DevOps tools you will likely encounter in the workplace

Agriad:


SophieHYe:
Thanks for the interesting proposal. Comparing DevOps tools is a very board topic. One kindly suggestion is maybe think about comparing the tools with a specific tasks or specific  criterion: effectiveness/performance/costs, etc.  We look forward reading your essay. I am now merging your PR.

fwallb:
Hi, I would like to give feedback on this essay when it is ready. PR for feedback proposal: #1054

PR #1030, opened 2021-03-29
Course automation submissions

heeenkie:


SophieHYe:
Thanks for the submission. Could you please also point out where the report of demos could be checked? 
We kindly advise you update the ReadMe and provide more information. Thanks!

heeenkie:
> Thanks for the submission. Could you please also point out where the report of demos could be checked?
> We kindly advise you update the ReadMe and provide more information. Thanks!

Thanks @SophieHYe, it should be fixed now.

SophieHYe:
Thanks.

PR #1067, opened 2021-03-30
Executable tutorial proposal: setup systemd-nspawn container on linux

8ar10der:
Member
Sui Yuehao yuehao@kth.se

### DIFF
--- a/contributions/executable-tutorial/yuehao/README.md
+++ b/contributions/executable-tutorial/yuehao/README.md
@@ -1,0 +1,15 @@
+# Executable-tutorial: Creat and config a systemd-nspawn container in your linux.
+
+## Members
+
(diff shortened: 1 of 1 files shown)


khesoem:
Hello,

Your chosen topic is relevant. However, if I understand correctly, your tutorial will be very short (probably with less than 10 commands). Won't it?

8ar10der:
> Hello,
> 
> Your chosen topic is relevant. However, if I understand correctly, your tutorial will be very short (probably with less than 10 commands). Won't it?

Not just create, I will introduct lots of configs, such as Resource control, Networking etc. It of course will more than 10 commands.

khesoem:
> I will introduct lots of configs, such as Resource control, Networking etc

In that case, please be more specific in your proposal. Determine what you will exactly show.

8ar10der:
Ok, I will make a new PR later.

PR #1066, opened 2021-03-30
Added myself (Fredrik) as participant to a project

fbjorkman:
I have spoken to Christopher and we will do the project together. In this PR both the folder name and the readme is updated.

PR #1090, opened 2021-03-31
Tutorial: Added team member and changed to locally run tutorial

ChristopherGustafson:
This PR is to add Fredrik Björkman as a team member for this executable tutorial.

It is also to change the tutorial slightly. We decided to have a tutorial that should be run locally instead of on Katacoda. This is because the resulting pipeline would be beneficial to have on the local machine and that it will integrate some other platforms like GitHub and Docker Hub.


ChristopherGustafson:
Adding Fredrik was also done in [this PR](https://github.com/KTH/devops-course/pull/1066) but we decided it was easiest to have all changes done using the same fork.

khesoem:
Hi @ChristopherGustafson 

I think you should remove [your old proposal folder](https://github.com/KTH/devops-course/tree/2021/contributions/executable-tutorial/chrigu) as well. Also please update the title to state your registered task more clearly.

ChristopherGustafson:
Hi @khaes-kth 

[This commit](https://github.com/KTH/devops-course/pull/1090/commits/24071e5375822d421af71ab8a4a525f32eb275a5) changes the name of my previous folder to include Fredriks kth-tag as well, so the old folder would not remain and I can therefore not delete it in the pull request.

As for the title, I am not sure which title you are referring to. I have changed the title of the PR to state what the tutorial is about, let me know if it is something else that is unclear.

PR #1094, opened 2021-03-31
Executable tutorial: Secure an API endpoint using Keycloak running in…

dmariel:
# Executable tutorial: Secure an API endpoint using Keycloak running in Docker
 
## Members
- [Dina Lerjevik](https://github.com/dmariel) (lerjevik@kth.se)
- [William Skagerström](https://github.com/wska) (wska@kth.se)

## Proposal
We would like to publish a tutorial on how to secure an API endpoint using Keycloak running in Docker and hosting it, on Katacoda. 

The aim of our tutorial is to:

* Provide and explain a simple API (likely a basic API using Python and Flask)
* Educate about security, Oauth2, OpenID Connect and Keycloak
* Showcase how to set up Keycloak and host it using Docker
* Go through how to set up a realm and add users with different levels of security clearance
* Use the existing Keycloak realm to secure the previously introduced API, which will now require access tokens and the relevant security clearance
 
We found an existing tutorial on Katacoda on this topic, however we believe that this tutorial was very basic since it merely only covers how to obtain an access token and using it for a request, which is only a few steps of the whole process.

### DIFF
--- a/contributions/executable-tutorial/lerjevik-wska/README.md
+++ b/contributions/executable-tutorial/lerjevik-wska/README.md
@@ -1,0 +1,18 @@
+# Executable tutorial: Secure an API endpoint using Keycloak running in Docker
+ 
+## Members
+- [Dina Lerjevik](https://github.com/dmariel) (lerjevik@kth.se)
+- [William Skagerström](https://github.com/wska) (wska@kth.se)
+
+## Proposal
+We would like to publish a tutorial on how to secure an API endpoint using Keycloak running in Docker and hosting it on Katacoda. 
+
+The aim of our tutorial is to:
+
+* Provide and explain a simple API (likely a basic API using Python and Flask)
+* Educate about security, Oauth2, OpenID Connect and Keycloak
+* Showcase how to set up Keycloak and host it using Docker
+* Go through how to set up a realm and add users with different levels of security clearance
+* Use the existing Keycloak realm to secure the previously introduced API, which will now require access tokens and the relevant security clearance
+ 
+We found an existing tutorial on Katacoda on this topic, however we believe that this tutorial was very basic since it merely only covers how to obtain an access token and using it for an request, which is only a few steps of the whole process. 


khesoem:
Good proposal. Good luck!

wska:
@khaes-kth 

Hi!

A while after starting to code and collect material we noticed another interactive tutorial that is pretty similar to ours but sets up and secures a NodeJS api endpoint.

The three major distinctions between that one and ours are that we plan on setting it up using an image using Docker/Kubernetes, instead of launching it through the use of a standalone application. We also planned on showcasing this for a Python API Endpoint (Flask) instead of a NodeJS one. Additionally, we are also showing this for both an API endpoint access via a token, but also how to redirect users to login via Keycloak and then redirect them back (versus for only showing how to get an access token and use it). When making Keycloak work for a redirectable endpoint, there are also some differences in how to set up the configuration of the sever, meaning that the other tutorial is not really "translatable" in its current shape for our usage area. 

While likely a bit too similar for distinction, we were wondering if it would still be possible to obtain a criterion of originality here? If no, should we perhaps focus on the other grading criteria? We feel somewhat confident in the other criteria except maybe "The tutorial is successful (attracts comments and success" which is really... difficult to plan out. Could you perhaps even clarify how that grading criterion works?

Thanks.

khesoem:
Hi @wska 

> if it would still be possible to obtain a criterion of originality here?

As far as I know, originality is considered related to other material on the web, not related to other tutorials in this course. However, if you think your tutorial is not original compared to other tutorials in this course, it is likely that your tutorial is not original on the internet as well. I would suggest that you focus and go as deep as you can when you are explaining the unique features of your tutorial. We should see the final result to decide about its originality.

> Could you perhaps even clarify how that grading criterion works?

AFAIK, this criterion refers to comments and discussions around your tutorial, especially on your submission PR. If TAs or other students are interested in your tutorial and engage in a discussion, it means your tutorial is successful. I do not think TAs will be too strict on this criterion, meaning that it is not required to have a very long discussion to get the point for this requirement.

wska:
@khaes-kth 

Ok! It was compared to other material on the web. But as mentioned, there are a couple of notable differences in their application area and what we are trying to accomplish with ours. Should we perhaps mention them in the submission PR and talk a bit about it?

I see. Well, that is extremely hard to "aim for" since it depends entirely on the other students and the TAs, so I guess the other ones are more defined if one wishes to aim for specific grading criterias. Thank you for the information.

khesoem:
> Should we perhaps mention them in the submission PR and talk a bit about it?

Sure, that definitely helps.

PR #1105, opened 2021-04-01
Course Automation submission

Spycsh:
You can view the concrete implementation code and the usage of this automation tool with the link below:

https://github.com/Chen-Zidi/check-modified-contents-link-validity-action

This project creates GitHub action, which checks the validity of the links in the modified content for a pull request. If there are invalid links in the added content, the output of the GitHub action should be false, and the pull request will be closed. Meanwhile, the invalid links are commented in the closed pull request so contributers can be notified about the invalid links. This action should be triggered by pull_request_target event, which means it can serve as an early-check before manual merging. Notice that we only check the links which start with http:// and https://


Spycsh:
conflicts with the folder name

PR #1116, opened 2021-04-03
Executable tutorial: CI workflow using GitHub actions for Node.js Express app, including automated testing and linting using Jest and Eslint

andnil5:
# Executable tutorial: CI workflow using GitHub actions for Node.js Express app, including automated testing and linting using Jest and Eslint

## Members
Anders Nilsson (andnil5@kth.se) Github: [andnil5](https://github.com/andnil5)

Niklas Wessman (nwessman@kth.se) Github: [nwessman](https://github.com/nwessman)

## Proposal
We will create a [Katacoda](https://www.katacoda.com/) executable tutorial including:
* A short introduction to automated testing with Jest, Linting with Eslint, and CI using GitHub actions.
* How to write mocha tests with Jest to test API endpoints on a simple Express application.
* How to set up a simple Eslint linter.
* How to create a CI workflow with GitHub actions, including automated tests and linting.

PR #1123, opened 2021-04-05
Tutorial: Static code analysis using Github Actions and Linters

jhammarstedt:
Proposed tutorial, will be created in katacoda.


khesoem:
Hi @jhammarstedt 

Your proposal is relevant, but it seems to be duplicated. See https://github.com/KTH/devops-course/pull/1116.

PR #1032, opened 2021-03-29
Essay proposal: BDD in DevOps

ghost:
# Essay Proposal 
I want to write an essay about why to (and how to) use BDD(Behaviour Driven development) in DevOps.
The essay will cover:
- DevOps testing (The testing hourglass)
- What BDD is (short on how it evolved)
- How BDD can improve devOps projects 
- The 3 amigos (who should use BDD)
- BDD tools and practices to use and follow

## Members
Agnes Petäjävaara <agnespet@kth.se> [anorangesky](https://github.com/anorangesky)

PR #1046, opened 2021-03-30
Proposal: feedback on BDD in DevOps

heeenkie:
Addressed PR: #1032

PR #1126, opened 2021-04-05
Feedback proposal on "Essay: BDD in DevOps (#1032)"

fbjorkman:


Deee92:
Hi @fbjorkman, the feedback for the essay proposed in #1032 is already being handled by #1046. Please create a feedback proposal for another essay or executable tutorial.

PR #1136, opened 2021-04-06
Essay proposal: Git - why is it so hard? And how to "Git gud"

christian-stj:
This is an essay proposal for an essay on the usability of Git.


khesoem:
Hi @christian-stj 

Git surely stands at the core of DevOps process. However, I think your proposal is too broad. You should make it more narrow and more specifically defined.

PR #1084, opened 2021-03-31
Essay: ITOPS VS DEVOPS: What is the difference and how they can be used together?

IC-kth:

PR #1159, opened 2021-04-08
Feedback proposal: Essay - ITOPS VS DEVOPS: What is the difference and how they can be used together?

LaraRos:
Feedback on #1084


khesoem:
Please add a comment to #1084 and note you would like to give feedback on it.

LaraRos:
Done!

PR #1165, opened 2021-04-09
Essay proposal

christian-stj:
Proposal for essay with outline added 

@khaes-kth I opened a new pull request as I could not update the one you closed

### DIFF
--- a/contributions/essay/cjgst/README.md
+++ b/contributions/essay/cjgst/README.md
@@ -1,0 +1,19 @@
+# Git - why is it so hard? And how to "Git gud"
+
+The goal of this essay is to dive deeper in what makes Git (and GitHub) hard. The essay will begin with some anecdotes and bring up examples of issues that newcomers have with Git and the complexity and unintuitiveness of it.
+
+The essay will hopefully have a humorous/ light take and be interesting and readable.
+
+## Author
+
+Christian Stjernberg - cjgst@kth.se
+
+GitHub: [christian-stj](https://github.com/christian-stj)
+
+## Relevance
+
+Git is a tool that is used in almost all software projects and is thus very relevant to DevOps. Usability of tools is also relevant to the workflow, which is integral to the DevOps process.
+
+## Innovativity/ Originality
+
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @christian-stj ,

Thanks for the update. I'm still concerned that your proposal is a bit too broad. Let us ask other TAs' opinion.

@Deee92 and @SophieHYe : What do you think about this essay proposal?

christian-stj:
@Deee92 @SophieHYe

Deee92:
Hi @christian-stj, I agree with @khaes-kth that the proposal might be too broad, but if you can ensure that you target very specific, technical aspects that make Git extremely relevant for DevOps, albeit potentially challenging for beginners, I think it would make for a meaningful contribution.

christian-stj:
@Deee92 @khaes-kth  I can try to angle the essay to be more about the usage of version control and its gain to development productivity and how improving on Git can make for a more productive development process. Which will make the essay more related to DevOps.

khesoem:
If I were you, I would change the topic and focus on a specific tool. "usage of version control" is too general and it is very hard to come up with an exciting essay on it.

christian-stj:
I am planning on writing about Git so it is a very specific tool. I will discuss Git and also Git servers which are very relevant to DevOps and the DevOps process. There will be background presented so the user can understand the relevance to the DevOps field and get some understanding of Git and then the essay will be focused on user interaction and concepts of the Git architecture that make it hard for new users to understand. 

I think this topic definitely should be relevant enough to the course. I will make sure that the essay will not be too broad, it will focus on the Git tool, its architecture and usability.

khesoem:
Fine.

Isn't your proposal very similar to this one? https://github.com/KTH/devops-course/pull/1092

christian-stj:
It does have some similar content, mostly the part explaining the structure of Git. Perhaps I should give up on this topic since it has met a lot of resistance.....

khesoem:
> Perhaps I should give up on this topic since it has met a lot of resistance.....

It's your call. I really do not want to resist too much, the sooner your PR is merged, the sooner I can proceed to assessing the next PR :-) However, we have to follow the rules of the course to the extent that we can.

christian-stj:
I will do a presentation instead, @khaes-kth thanks for the quick feedback.

PR #1182, opened 2021-04-12
Submission: Demo 🎥

annsudo:


Deee92:
Hi @annsudo, thanks for your submission. I see that your video is currently 3 minutes and 8 seconds long. I'm merging, but try to shorten it to be exactly 3 minutes before the demo session in week 7.

annsudo:
@Deee92 thank you for the tips, will fix!

annsudo:
@Deee92 Done! #1260

Atema:
Thanks for a demo on this interesting topic; I wish I had known about it when working on my course automation action! The introduction clearly describes the tool and the same issues that I had. However, I would remove the video of the `act` repository in the background (0:27 to 0:45); it feels a bit distracting because I wasn't sure whether I should focus on it.

For the installation, I think that many people are not familiar with Homebrew. Maybe you could shortly say what `brew` is, or use the general installation command `curl https://raw.githubusercontent.com/nektos/act/master/install.sh | sudo bash` instead.

The next part shows many features of act, which is nice, but feels a bit like an information overdose. However, the last (demo) part clears things up a lot. I don't know if it would be feasible (it would require quite a bit of editing), but I think the content overall would be clearer when the two parts are combined into one, explaining the flags along the way while using them.

Overall, nice job, and thanks for teaching me about this tool!

vladomit:
**Strengths:**
Really interesting topic,  I wish I knew about it before !
Clear easy to follow

**Improvement:**
Is this also working on linux ? Just mention it I would say

Very good tutorial ! Good job !

yuxinmmmm:
# Feedback
## Overall
It's a nice video. It would be helpful to those who are interested in how to run the Github actions locally with Nektos Act and Docker.
## Good points
- Perfectly on time of 3 mins.
- At the proper speed and a clear voice.
- Actionable takeaway messages.
## Suggestions
- Could reduce the volume of background music.

Agriad:
# Feedback
### author: arieltan@kth.se

## Positive
* Nice animation  
The animation enchances the style of the vidoe and makes it pleasing to watch.

* Nice flow  
The flow of the video is nice and the pacing is great.

* Explanation and showcase  
I really like how you explain first about the topic and then follow it up with a showcase of it.

## Improvements
* Handling secrets title page  
The title page for this section was a bit too fast so maybe it could be slowed down.

* Description  
I notice that there are timestamps in the description. Perhaps you can make them into Youtube chapter timestamps so it also shows in the video timeline.

johanmallobakken:
# Feedback

## Strengths

Wow, I'm very impressed by this. Overall super solid submission. Great job!!! 

## Improvements (if any)
You could maybe mention something for other operating systems, or specify what OS you are running when you run the brew install command. 

But this should be good enough for distinction already :)

hengque:
Nice demo, definitely something I would have had use for when working with GitHub actions in this course!
+ Overall good pacing: not rushed and not too slow
+ Good video description and nice inclusion of timestamps (though even better if they were clickable ;) )
+ Nice background music that never stole focus from the content

- Would have been good with a higher resolution; for instance the editor-views very a bit blurry
- Some places where a bit too quick, like by 2.20 when you mention something went wrong, it doesn't highlight this and only shows it for a short period of time

PR #1194, opened 2021-04-13
Email fix in readme

jhammarstedt:
Email was wrong so updated it


jhammarstedt:
@Deee92 Thanks for pointing it out, now it's fixed!:)

PR #1196, opened 2021-04-13
Executable Tutorial: Building a CI/CD Pipeline for a small frontend in a Kubernetes Cluster using Terraform

Internet-Person-IP:
# Executable Tutorial: Building a CI/CD Pipeline for a small frontend in a Kubernetes Cluster using Terraform

## Members

Taqui - taqui@kth.se
GitHub: [Internet-Person-IP](https://github.com/Internet-Person-IP)

## Proposal
I want to create an Executable Tutorial for creating a CI/CD pipeline for a Kubernetes Cluster using Terraform. The frontend will be super simple with a counter and click function.

The current structure of the tutorial will be creating a terraform template that does this:

- Creating the initial frontend, set up frontend with simple tests
- Creating a small cluster
- Containerizing the frontend
- Deploying the frontend to a small pod
- Creating the CI/CD pipeline
- Make changes to the frontend and actually push it to master
- Check that the CI/CD pipeline works correctly and then make sure that the deployed website works correctly

PR #1225, opened 2021-04-15
Feedback on #1196

thestar19:
I would like to give feedback on #1196


khesoem:
Thanks. Let's wait until #1196 gets merged.

PR #1232, opened 2021-04-16
Presentation Proposal: Cloud and the journey for a start-up to Fortune 500 (Follow up)

MDRezaulHasan:


khesoem:
Hi @MDRezaulHasan 

Sorry, but I think we cannot approve this PR. It is similar to the previous one.

MDRezaulHasan:
Hi @khaes-kth, Martin said us to follow up with the previous presentation. We can focus in detail on load balancing and auto-scaling. So what we did, we make a new presentation that explains deep of auto-scaling and load balancing. If you want  I can send you the email which marting send us.

PR #1240, opened 2021-04-16
exectuable toturial proposal: deploy php web application using docker

Chen-Zidi:
Hi! Our team had an executable toturial proposal about teamcity in docker a few weeks ago, but we found that is hard to make it run on katacoda. katacoda seems not support this environment. We had an lab session, but we still got no progress on this topic. Our motivation is to make our toturial run on katacoda, so we think the best solution is to change the topic. Here is our new topic. If it is not possible to change the topic, please comment me! Thanks!!

# Executable-tutorial: deploy php web application using docker

## Members

Chen, Zidi: zidi@kth.se, https://github.com/Chen-Zidi

Chen, Sihan: sihanc@kth.se, https://github.com/Spycsh

## Proposal

We want to write a toturial on katacoda. 
The idea is about to deploy a php web application using docker. 
Docker is a useful too for CI/CD, which should fit the purpose of the course. 
Since there is no such toturial on katacoda, we want to help to improve.

If there are more suggestions, 
please tell me :)


khaes-kth:
Hi @Chen-Zidi 

Unfortunately, it is not possible to change the topic of your tutorial. You can change the technology (Katacoda), but not the topic.

Chen-Zidi:
We cannot find other platforms like katacoda. mybinder does not support our toturial. If we write an article, we can not get remarkable. We have also wrote emails to a few teaching assistants, but there was no reply. Could you please give us some advice like other possible platforms or who can help with our problem?

PR #1252, opened 2021-04-19
Presentation proposal: Gradle vs. Maven

majate:
# Presentation: Gradle vs. Maven
## Members

Maja Tennander (majate@kth.se) [Github](https://github.com/majate)

Axel Elmarsson (axelelm@kth.se) [Github](https://github.com/elmaxe)

## Topic
We want to talk about the differences between Maven and Gradle.

### Outline
- Short introduction about package managers and their role in DevOps
- What is Maven and Gradle?
- Pros and cons
- Which should be used for which projects?


khesoem:
Good luck!

PR #1261, opened 2021-04-19
Withdraw proposal #1201

vladomit:
Withdraw proposal #1201 since my teammate @dsharifi is discontinuing the course.

I am joining another task  #1189, changes will be made soon by @ysndr 

Thanks

PR #1262, opened 2021-04-19
Amend students involved in tutorial

ysndr:
Daniel had to drop the DevOps course. Instead I ask to amend the approved proposal to do a tutorial with Vlado who withdraws from his previous proposal in #1261

### DIFF
--- a/contributions/executable-tutorial/yannik-vladom/README.md
+++ b/contributions/executable-tutorial/yannik-vladom/README.md
@@ -23,4 +23,4 @@ Link to the tutorial: *TBA*
 ## Authors
 
 - Yannik Sander ([yannik@kth.se](mailto:yannik@kth.se)), [ysndr](https://github.com/ysndr)
-- Daniel Sharifi ([dsharifi@kth.se](mailto:dsharifi@kth.se)), [dsharifi](https://github.com/dsharifi)
+- Vlado Mitrovic ([vladom@kth.se](mailto:vladom@kth.se)), [vladomitrovic](https://github.com/vladomitrovic)


khesoem:
Hi @ysndr ,

Unfortunately, there is no withdrawal option for proposals.

If Vlado still wants to join you, we can approve this change.

ysndr:
This is not a withdrawal request but a change to accommodate both me and Vlado to do a task together now that our common partner Daniel had to discontinue the course.

I am sure @vladomitrovic will approve joining this task!

khesoem:
> This is not a withdrawal request

If I understand correctly, Vlado wanted to withdraw his proposal, and join this one. I just wanted to note that there is no withdrawal option. Considering this, if Vlado still wants to join this task, we can approve it.

ysndr:
> Considering this, if Vlado still wants to join this task, we can approve it.
it would be his 5th task to participate in.

In either case I want to comment that we cannot withdraw from a task even in case the preconditions of a task change unexpectably (i.e. a  member dropping the course). It's also somewhat frustrating that while student tasks cannot change the rules to this course are repeatedly being restricted without an explanation why so. I don't mean to be offensive but ask for a dialogue and more transparency.
@khaes-kth, @bbaudry @monperrus

khesoem:
> it would be his 5th task to participate in.

If this is the case, I think he cannot join you in this task.

> It's also somewhat frustrating that while student tasks cannot change the rules to this course are repeatedly being restricted without an explanation why so

I think one reason is that other people may have registered feedback for a task; therefore, if the task is withdrawn, it will be unfair to feedback givers. To have a general rule, we say that once a proposal is merged, it cannot be withdrawn.

monperrus:
> Considering this, if Vlado still wants to join this task, we can approve it.

Fully agree with @khaes-kth, @vladomitrovic can replace Daniel.

> repeatedly being restricted without an explanation

Nobody is meant to be restricted. We only make sure that we all have the same understanding about the work to be done, before it is actually done (topic, team)

vladomit:
>  think one reason is that other people may have registered feedback for a task

I understand, but in this case the feedback giver cancel his PR. I also think it's unfair to be stuck with a task without having the possibility to withdraw it in special cases such as teammate discontinuing the course.

vladomit:
> Nobody is meant to be restricted. We only make sure that we all have the same understanding about the work to be done, before it is actually done (topic, team)

I totally agree, the only issue here is the fact that I can't withdraw my essay proposal (because my teammate dropped the course) to join this task

khesoem:
Hi @vladomitrovic 

Martin and I discussed this issue. We believe it makes sense to let you move from your previous registered task to this one if:

1. No one has registered to give feedback on your previous task, and
2. You have already registered four tasks.

If you meet these conditions, please create a PR to withdraw your previously registered task again. Also, add a comment in your previously merged PR stating that PR is withdrawn.

ysndr:
@khaes-kth I'm happy this could be resolved! 💯

vladomit:
> If you meet these conditions, please create a PR to withdraw your previously registered task again. Also, add a comment in your previously merged PR stating that PR is withdrawn.

Yes I am meeting these requirements. I am sending the new PR

I'm happy we resolved this, thank you !

khesoem:
Good luck with this tutorial!

PR #1314, opened 2021-04-22
Exacutable tutorial propsal: setting up splunk in docker container

Augustjon:
Here is a new proposal for a tutorial as the previous was to similar to an older PR. Hope this will suffice :) @khaes-kth


khesoem:
Good luck!

PR #1364, opened 2021-04-26
Remove Periods in Word Count

felix-seifert:
* Enhance command for counting word with removal of periods before counting
* LaTeX files often use period signs to seperate names of sections/subsections from their page number in the content table. To avoid these periods being counted as words, periods should be removed before counting the word number.


khesoem:
Good improvement. Thanks!

PR #1367, opened 2021-04-26
Essay submission: ITOPS VS DEVOPS: What is the difference and how they can be used together?

IC-kth:
Initial [PR ID](https://github.com/KTH/devops-course/pull/1084)
Feedback by @LaraRos with [PR ID](https://github.com/KTH/devops-course/pull/1159)

Note:
Based on the feedback we receive from @LaraRos we will update our essay with the final version.


LaraRos:
Thanks, I'll get back with a review by tomorrow (hopefully)!

IC-kth:
@LaraRos  looking forward to it. Thank you! :)

LaraRos:
Hey! Here is the feedback: #1398

Please react to the feedback as well! I need the reaction for one point :)

Regards, Lara

LaraRos:
Here is the feedback as a comment:

## Overall
* The paper discusses ITOps and how it is different and similar to DevOps. The authors also mention how these set of practices manifest in a company, and where they might cause issues for the company.
* At the moment, the word count is too small, but this is expected to change for their final submission. 
* The references are okay, mostly deriving from blog posts. Though one would prefer more scientific articles as references, I think the authors have done the best they could with this topic.
## Title
* The points directly to the subject and is relevant to the essay.
## Introduction
* Intro discusses current IT infrastructures, and the role of ITOps in such an infrastructure
* Grammatical error: _Due to this dependency, arose [...]_
* Currently, the intro is short and precise. Perhaps there is a too large focus on ITOps, there could be more of a discussion regarding both DevOps and ITOps in broader terms. The essay is about the relationship between these two practices, so I think a more even split would be apropriate.

## ITOps
* This section delves deeper into ITOps, by discussing how it may be used in a company or organisation.
* Indeed, ITOps is quite broad, however I as a reader would like more in depth examples of what ITOps looks like in an organisation. Currently, the tasks related to ITOps that are mentoned are network infrastructure, computer operations and help desk, and management of service and device. Later, the authors give some examples of what this could entail: _Examples of some of these functions can include providing users with remote access using VPN or two­factor authentication_. I think this is too broad. How does ITOps provide the users with this? Which users are we talking about? I think one could go into the details of this.

## Differences between ITOps and DevOps
* This section focuses on the differences between the two practices. The conclusion is that DevOps is more agile and adventerous, whereas ITOps is more stable and less prone to failure. 
* I think giving the figure in this section a caption would be good (actually all figures should have this), because right now it seems quite sudden and a reader may not be able to fully digest what the authors are trying to say with it.
* I also think you could incorporate some of the differences mentioned in the figure in the essay as well. You could take about how DevOps os more decentralised while ITOps is more centralised, for example.

## The Ways ITOps and DevOps Work Together
* In this section, we learn how the two practices might come together in a company. They focus on that communication between the two teams is essential in an organisation
* One issue that can occur between the DevOps and ITOps that the essay mentions is change: _The problem that we would like to highlight is related to an inevitable and frequently occurring phenomenon in software development, namely, change._ Could this be elaborated more? Perhaps discuss what type of change that can cause this? I feel like just _change_ could entail anything to be honest. I'd like some more specifics, preferable relatable to tech.
* One sentence is very long and hard to understand, I'd suggest breaking it down: _More specifically, when both teams are provided with real­time, comprehensive, and accurate data which is extracted, under a standard model, from a data pool, a foundation for decision making is created._
* Upon reading what ITOps was about, I felt like DevOps and ITOps might engage with each other much more, even using the same software. For example, a company might develop a two factor authentication system internally for the company, which ITOps then maintains to enable users in the company to use this service. Is this the case? Or does ITOps and DevOps generally never engage in matters which use the same software? I think one could discuss how DevOps develops something which then ITOps maintains in their work (if this is how the two practices work together).

## Common Tools for ITOps and DevOps
* This section focuses on the tools the practices use, and mentions that they should use the same tools to maintain a good cooperation between the two teams
* The figure reference is wrong here
* The figure in general is not really discussed, it's just a table with some tools. I'd like the authors to go through it a bit more and highlight some noticeable tools. Perhaps a caption would do great here.

## Conclusions
* I think the conclusion is rushed and doesn't relate to the essay, as highlighted by the short mention of AIOps. As readers, we barely know what AIOps is about, and it seems like this was just thrown in the essay. Perhaps a sentence or two about how it replaces DevOps and ITOps would be apropriate. Indeed, one could even talk about what AIOps is in a separate section if apropriate.
* In general, I think the conclusion should tie together the subject of the essay, which is comparing DevOps with ITOps. A sentence or two about their differences and similarities would be good, and then perhaps one could finish by mentioning AIOps.

## Some general comments
* I think the essay is quite broad when discussing some terms. I'm not sure if this should be changed considering the scope of this essay. Personally, I'd like more clarity regarding ITOps at least, I feel like we were barely brushing on the surface.
* I think you should use figure captions on all of your figures. I also think you should refer to them more often.
* Perhaps more scientific sources would be good to use in your report. Here are some that I found with Google Scholar, perhaps they are relevant to you? [first link](https://link.springer.com/chapter/10.1007/978-3-030-64849-7_28) [second link](https://www.sciencedirect.com/science/article/abs/pii/S0098135498001069).

IC-kth:
I will close this pull request and create a new one for the final submission.

PR #1389, opened 2021-04-27
Final Submission - Executable Tutorial: Implementing AWS Lambda with AWS SES and invoking it with HTTP Request

sfkwww:
# Tutorial Submission: Implementing AWS Lambda with AWS SES and invoking it with HTTP Request
## Members
- [Sebastian Williams](https://github.com/sfkwww) (swill@kth.se)

## Proposal

I would like to create a tutorial on how to setup the serverless microservice AWS Lambda. The example used in the tutorial will be triggered by AWS SES and then send a response to a specific email.

The main steps of the tutorial will be as follows:
- Setting up SES
- Creating the Lambda function
- Creating an API Gateway
- Configuring the permissions for the Lambda

Microservices such as AWS Lambda are commonly used within DevOps to build a single application as a set of small services.

## Submission

Due to invoking AWS Lambda with SES requires a domain I opted to use an HTTP request as the lambda trigger instead.

The Katacoda tutorial can be found [here](https://www.katacoda.com/sfkwww/scenarios/aws-lambda-ses) and the repo can be found [here](https://github.com/sfkwww/katacoda-scenarios).

## Changes based on Feedback

- Added an example scenario of why using Lambda like this could be useful in the intro.
- Rewrote parts of step 2 to make it more clear.
- Add an explanation on how to send emails to unverified email addresses.
- Added some more details to step 4 to reduce confusion.
- Added background information on what Lambda Triggers are.
- Added a link for instructions on invoking AWS Lambda on received emails.

### DIFF
--- a/contributions/executable-tutorial/swill/README.md
+++ b/contributions/executable-tutorial/swill/README.md
@@ -14,10 +14,19 @@ The main steps of the tutorial will be as follows:
 
 Microservices such as AWS Lambda are commonly used within DevOps to build a single application as a set of small services.
 
-## Submission for Feedback
+## Submission
 
 Due to invoking AWS Lambda with SES requires a domain I opted to use an HTTP request as the lambda trigger instead.
 
 The Katacoda tutorial can be found [here](https://www.katacoda.com/sfkwww/scenarios/aws-lambda-ses) and the repo can be found [here](https://github.com/sfkwww/katacoda-scenarios).
 
 I hope to receive feedback from @oskstr as per #1164
+
+## Changes based on Feedback
+
+- Added an example scenario of why using Lambda like this could be useful in the intro.
+- Rewrote parts of step 2 to make it more clear.
+- Add explanation on how to send emails to unverified email addresses.
+- Added some more details to step 4 to reduce confusion.
+- Added background information on what Lambda Triggers are.
+- Added a link for instructions on invoking AWS Lambda on received emails.
--- a/contributions/executable-tutorial/swill/README.md
+++ b/contributions/executable-tutorial/swill/README.md
@@ -1,4 +1,4 @@
-# Tutorial: Implementing AWS Lambda and invoking it with AWS SES
+# Tutorial Submission: Implementing AWS Lambda with AWS SES and invoking it with HTTP Request
 ## Members
 - [Sebastian Williams](https://github.com/sfkwww) (swill@kth.se)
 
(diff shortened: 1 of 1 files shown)


oskstr:
## Feedback

### Overview
It was a very nice tutorial. I was somewhat familiar with the subject before so navigating AWS's confusing user interface wasn't too bad for me. But that is hardly something I can blame you for anyway. The pictures helped a lot with that as well.

I like that the steps are very concise and easy to follow. 

According to the Katacoda Design Guide, a scenario should have the form of a "loosely defined story" to create some emotional response making the learning more likely to stick. So maybe "why" we are learning this or at least what we are trying to do is worth adding.

I think the length of the scenario was good.

### Intro
Following the tutorial I realise that it is a homage to [NARKOZ/hacker-scripts](https://github.com/NARKOZ/hacker-scripts), in particular `kumar-asshole.sh` (funny easter egg btw). Knowing that I can see why having problems with "receiving emails and processing them" is an issue. But since we don't have the background at this stage that doesn't really matter.

Maybe have some more context here. If you go the story route you could have a scenario where we are trying to solve a particular problem. For example having someone bothering us via email with a trivial task over and over again - and we want to automate that. For that we can use a Lambda function and why we would choose that.

### Step 1
Very simple step. Easy to follow. I already had an AWS account which made this trivial. I just had to log in. I didn't bother to create another account for the purposes of this feedback, I'm sure you'll understand.

### Step 2
Most of it was straightforward except for 
> "...create a new role for `Lambda`..." 

as AWS calls it a Use Case. So you could perhaps clarify that with something along the lines of 

> "... create a new Role and select the `Lambda` use case. Then add the `WriteEmails` permission policy ... " 

You can probably think of some better wording. 

Between the permissions page and page where you give the role a name there is a "Tag" page. I don't know if it is necessary to tell people to skip that page to "Review" (where you name the Role). Something to think about perhaps. I was a bit confused at least, but that is more of an issue with AWS than with your tutorial.

### Step 3
Good thing that you have a short explanation of SES again in case the user didn't read the intro thoroughly :+1:. 

Even though it is easy to add two email addresses you can simplify this step a bit by just sending and receiving using the same address. I would also like to see some more explanation here. The reason why you have two is because you need to verify addresses you can send to as well, right? How do I send to any email address?

### Step 4
Again, a fairly straightforward step. You might want to add a word or two about which runtime to use (even though we are using the default). That we are going with Node.js because we are doing a Lambda in Javascript. But you could technically choose from a bunch of languages here.

I also found setting the permission in AWS here a bit confusing. You could explain that with a few more words. For example:

> "... by pressing `Change default execution role` and then pick `Use an existing role` and select `LambdaEmail`... "

You could also explain with a few words what the code is doing and what you might have to change if you have used other email addresses than the ones provided.

### Step 5
Good explanation and good troubleshooting step. 

### Step 6
Any reason why we are going with an `HTTP API` instead of a `REST API`?

I would also like to see some more information about what a trigger is. 

If I would like to trigger it based on incoming emails, how would I do that? If you go with explaining that is what you ultimately would like to do. Maybe point me in the right direction at least.

sfkwww:
Thanks for the review, have implemented your suggested changes!

PR #1419, opened 2021-04-29
Essay FINAL submission

agnesforsberg:
We are ready for feedback @Chen-Zidi and @amaothree, looking forward to your feedback!


Chen-Zidi:
# Feedback

If you feel anything unclear in our feedback, just contact zidi@kth.se and yuehao@kth.se! :blush: Looking forward for you revised version!

 ## Overall

This is a great essay with a lot of information, keeping the word count still under 2000. The structure is reasonable and clear, including all necessary parts, with concise and direct section titles. In addition, the content is detailed, so that this essay is self-contained to readers. It also contains 10 references which should fit the reference requirement of grading criterion.

But, in our opinion, the essay can still be improved in the following aspects in general. The first thing is that sentences are sometimes not organized in suitable relationships. The usage of "but", "therefore" and other conjunction might not be so reasonable. In addition, the references are not placed well sometimes. There are places where you cite a paragraph/multiple sentences. Check [here](https://academia.stackexchange.com/questions/8804/how-to-add-citation-for-a-whole-paragraph) to see a discussion about how to cite. We personally recommend that you can simply do something like: **sentence 1. sentence 2. sentence 3. [1]**, as the popular answer in [here](https://www.researchgate.net/post/Do-I-have-to-cite-after-every-sentence-when-paraphrasing-a-whole-paragraph). Furthermore, maybe you can give more transitions between subsections and sections. But this is only a minor point. The title `A comprehensive comparison of DevOps and Site Reliability Engineering` is a little bit too ambitious from our point of view. Maybe you could delete `comprehensive`. But it depends on you.

Since most of your references are urls, we found a few free online books about SRE which might help: [Site Reliability Engineering](https://sre.google/sre-book/table-of-contents/) and [The Site Reliability Workbook](https://sre.google/workbook/table-of-contents/).

 ## 1 Introduction

The introduction is good, but it still can be improved. You explain why we need SRE and DevOps and what the essay wants to do, which gives reader a clear understanding about what they can expect. Although we had little idea about SRE, we felt really interested in your essay after reading the introduction.

Here are some suggestions:

We feel a little bit hard to understand the first paragraph. The words to connect between the sentences might not be suitable. If our understanding is correct, maybe it could be written like this:
 
 > "Efficiency... **In addition**, reliability is also highly critical for companies and software engineering. Lack of reliability could result in terrible consequences, which has the same impact as a failing service. The traditional way to improve reliability is to ask the development team to deploy as quickly as possible. **However**, the operation team usually would rather keep the safe, old code that is known to be reliable in deployment."

In this way, the first paragraph could connect to the second one in a more fluent way. We also recommend adding a comma after "However".


 ## 2 Background

 This section is really detailed. It provides a lot of background information to make outsiders understand what are DevOps and SRE. You offered a solid basis before comparing the two concepts. Another aspect that we like is that you make the important letters bold, which makes the essay more direct and clearer. 

 Here we give suggestions in subsections:

## 2.1 DevOps

- The second sentence is repeated in "done". Maybe this way is better (if our understanding is correct):

> The work, which traditionally is responsible by different teams, now tends to be done collaboratively and in a more synchronized way.

- The fourth sentence should start with an upper case. 

> "What the model accomplishes ..." 

- Maybe you need a comma before "while". 

> "new operational features, while at the same time maintaining the upkeep ..."

- We are not sure why "Mantra" needs upper case here.

- Here we offer an optional style to write the third concept Lean IT. So that you can avoid using "you" in scientific writing and reduce the complexity of the sentences a little bit. It would be better to avoid using second person pronoun, as stated [here](https://academicguides.waldenu.edu/writingcenter/scholarlyvoice/second).

> The third concept is **Lean IT**, which is a workflow that ensures to maximize the gain of consumers, and minimize time waste and bad workflow.

- We think that you might need to cite here.

>  DevOps is usually split into 4 main phases that surround application lifecycles or products. [10]

- Maybe use comma to split the sentences when you use conjunction words like "Once" and "however".

- Figure 1 is too far away from this section. Maybe place the figure just after this subsection would be better.

- Maybe you can make a conclusion after this subsection and flow to the introduction of SRE.

 ### 2.2 Site Reliability Engineering - SRE
  
- We find this sentence a little bit hard to understand. Could you please clarify it?
 
> Google describes SRE as operations when you treat it as a software
problem [6]. 

- As we said, maybe you can reference in this way to include the previous sentences. (We think the previous one should be included here.)

> ... and SRE at Google grew from that group. [6]

- The second paragraph might need a reference.

- We find it is strange to use `but instead also ...` in the second paragraph. Perhaps just delete `instead`.
- Not sure why `diversity` is in Italic.
- Make a transition to the comparison would be good.


 ## Comparison
This section is also clearly structured. Three subsections cover different good points of comparison, which also give us a deeper understanding on DevOps and SRE. You have a great figure to explain the relationship between DevOps and SRE.

Some suggestions:

### 3.1 Concepts and Philosophies

- Figure 2 is a really nice figure. May more descriptions.
- Please check from which sentence that you start using the content of [7]. Maybe using `. [7]` at the end of the paragraph. By the way, we think that this video is really nice and helps to understand the topic.

### 3.2 Workﬂow

- In the third sentence, you use two `effective`. Replacing `effective deliveries` by `quick deliveries` could be better.
- By the way, we found it is a great comparison here: DevOps aims to predict problems, but SRE aims to fix.
- Add punctuations here:
>  This leads to expensive costs if there is failure, therefore, as mentioned earlier, there is a heavy focus on not overlooking mistakes in the pre-deployment workﬂow [8].  

- Remember to add comma to split sentences when using "however".
- Not sure the meaning of the sentence. Do you mean (need) more efficient testing?
  
> This might take longer time and more efficient testing after deployment however heavily reduces the cost of failure.

- The last sentence has a misspelling `maintaing`.

### 3.3 Work culture

- The second sentence is not clear. Please clarify.

> , and while DevOps engineers exist it is not as common.

- Not sure why `invented` is in Italic.

- In the last sentence in the first paragraph in the subsection, you mentioned a book, maybe you need to figure out which book it is and give the name here.

## Conclusion

We think it is a great conclusion. It summarizes the essay in a succinct way with all necessary details. You have also pointed out an interesting thing: how people treat SRE and DevOps in jobs. Nice.

Maybe you can add some discussion for future research.

George-Bassilious:
Thank you for the feedback @Chen-Zidi and @amaothree ! We appreciate the tips and ideas that you presented us with. We have done revisions according to the feedback to make the essay more clear and understandable! Sadly we could not fit in some of the recommended parts that you told us would be nice to have because of the word limit! Hopefully our essay is improved now! Have a nice weekend ☺️

agnesforsberg:
We have now submitted the final essay, and are ready for grading! @cesarsotovalero

PR #1465, opened 2021-05-03
fix: remove unwanted workflows

khesoem:
To avoid confusion and headache with unwanted PR failures, this PR removes the GHAs that we are still not sure if we want to integration into the repo.


SophieHYe:
Thanks Khashayar for cleaning our repo. I am merging your PR now.

PR #1480, opened 2021-05-05
Demo: CI and CD pipeline using Azure DevOps

MDRezaulHasan:

PR #1485, opened 2021-05-06
Final Demo Submission: CI and CD pipeline using azure DevOps.

MDRezaulHasan:
Hi @khaes-kth, 
I create this new PR. And in here I only change one file. 
My Old PR: #1480

### DIFF
--- a/contributions/demo/mrhasa/README.md
+++ b/contributions/demo/mrhasa/README.md
@@ -7,7 +7,8 @@ GitHub: [MDRezaulHasan](https://github.com/MDRezaulHasan/)
 
 ### Proposal
 
-In this demo I am focusing on how to deploy an asp.net web application in azure devops. 
+In this demo I am focusing on how to deploy an asp.net web application in azure devops.
+
 ### Topics to covered
 
 1. Build an asp.net project and push it on azure devops
@@ -16,3 +17,4 @@ In this demo I am focusing on how to deploy an asp.net web application in azure 
 4. Explaining about yml file.
 5. Build a release pipeline for deployment.
 
+[Screencast](https://youtu.be/H9bfJu0A9WY)


khesoem:
Thanks @MDRezaulHasan

arxra:
# Feedback
The goal of the video was clear, and it does cover what one expects from the title.
However, the voice over was all over the place, lots of "ahh" and "eeerhm" which could easily be edited out, suggestion is to record the video and audio separately.

Some more visual feedback about the pipeline results would also be nice, something that does more than a single stage in the set up pipeline but not necessary to get the point across. But more information on multistage build pipelines would be a great use of the time saved by cutting out "uuh"s. 

All in all relativtly easy to follow demo, but I would probably have chosen a different tutorial after the first 20 seconds if I found this while trying to set up azure pipelines from the unfocused voiceover.

yuxinmmmm:
# Feedback
## Overall
It is a good video demo. With this video, it’s easy to follow the steps of the CI and CD pipeline using Azure DevOps. And the voice is clear, a little too fast but understandable.
## Suggestions
- Consider adding some Actionable takeaway messages.
- You can give some motivation and explain what is the Azure DevOps.

nwessman:
# Feedback
Interesting topic, it was a nice introduction to it! 

It sounds and feels like the voice over is done in one take. This makes it feel relaxed and it creates a good flow of voice in the video. At the same time, this includes your habits of saying “um” and similar which breaks the flow in the video and it makes the demo not sound as prepared as it could be. There is one part where you say “I think I have to..” which is not a very clear way to present the information. I think the video would benefit if you recorded it part for part to lose the problems with uncertainty.
 
This is a video, so you do can edit around the waiting times instead of cancelling processes live because of time constraints. Like at 1:30.

You should delete the `Ending Screen`  I think you added to your video since the last 20 seconds of the video has video recommendations and your profile picture all over which makes so I can not see the content of the demo instead.

Your workspace is also very cluttered with your Windows Activity Bar at the bottom and all your Favorites in Chrome. This makes the video a bit more cluttered with distractions than needed. You could at least remove the Windows Activity Bar to get rid of some of this.

Overall I think you lost a lot of time because of the way you recorded the video, and that you could get the message across more clearly if you changed the way you record it.

Chen-Zidi:
# Feedback

Good video. I feel like you are just sitting besides me and explain, which makes me relax and easy to follow you. From the video, every step is clearly visible to watcher, so there should be no misunderstanding about what to do, which is also good.

Here are some possible improvements. I personally think that the background information is not enough. You could speak a little bit faster, and maybe use a few slides to explain what is Azure and why Azure. In addition, the end of the video feels like a sudden stop. You can give a few bullet points at the end, working like taking home messages or a conclusion.

In all, thanks for the video! Azure devops is a new topic for me. This video makes me learn a lot.:yum:

johanmallobakken:
# Feedback

## Strengths

Very instructive and clear video! Straight to the point, clear speaking and interesting topic. 

## Improvements

You can add a take-home message. 
The video could be structured better.
The video could be more visually appealing.

In general a very good video! Great job! :)

m4reko:
Thank you for a nice demo! This is my feedback for you:

## Good points
- Super interesting topic
- Perfect scope of the subject 

## Points of improvement
- The introduction is very short and you go into technicalities very fast. More context in the intro would be awesome.
- You should be able to write down what you want to say and record it separate from the screen recording. This would give you more flow and it's pretty easy to sync the screen recording afterwards in any free video editing tool. Now it's a bit stuttery and hard to follow your reasoning and explaining.
- The outro could also be improved by adding more context, maybe a summary and a take-home message.

MDRezaulHasan:
Thank you, everyone. I update this demo based on feedback.

PR #1500, opened 2021-05-10
[Submission] Demo: CI&CD with Kubeflow Pipelines in GCP

sestys:
Final demo submission, more in the readme.


sestys:
@jhammarstedt

Sebberh:
Video and sound was very good, there were some slight inconsistencies but it didn't detract from the video. Everything on screen was readable and it was easy to follow what was happening,although pausing the video was sometimes needed.
The video could use a little more explanation and context to highlight why this is important and how it is different from other approaches. But hey, the time-limit is what it is.

cpiehl1:
Overall a very nice demo. The use case is motivated at the start and the end. It was nice to see an outline of the content at the start. I also liked that you introduced the structure of the pipeline before going into detail. Visually, the video was quite good, although maybe a bit too fast at times.

One concern is that the voice-over felt too fast, making it hard to understand what was being said. This was especially noticeable during the explanation of the pipeline. I had to rewatch this step several times to understand it.

The scope is very ambitious and it is clear you have done a lot of work, but maybe it is a bit too much to squeeze into a 3 minute demo. Perhaps just focusing on the kubeflow pipeline would have made it feel a bit less overwhelming.

patricia-naccachian:
The demo was very clear and excellently outlined. From a technical perspective, it explained very well the procedure of the deployment of Kubeflow Pipeline while remaining within the 3 minutes limit. As for the way the demo was constructed, the background music was not distractive, and the narrative was excellent. However, more discussion should have been taken on why it matters to DevOps.

carinawic:
The video was very well made. I found the introduction especially good because it provided a clear overview of what the video contained. I think you covered an interesting topic, and connected it back to why it matters to devops in a good way. The animation in the beginning was cool. My suggestion for feedback includes making sure that the subtitles correspond to the exact words that are being said. At 0:57, the audio says "cloudbuild", but the subtitles say "cloudbuild.yaml", which is slightly inaccurate. The pace of the video was a little bit too high, although this is difficult to change due to the strict time limit.

Spycsh:
**Overall good job guys! And I will give some feedback below**

**Positive**
* The video itself clearly show the whole flow of build a pipeline of Kuberflow for CI/CD of ML models and is itself tightly relevant to MLOps
* There is a relevant GitHub repository with an attractive README 
* The voice is clear and the backgroud music fits well

**Improvements**
* Maybe a whole architecture image can be added to simply demonstrate the whole work flow because this project seems to include a lot of tools

PR #1493, opened 2021-05-10
Video Demo Submission: GitHub Integration with Jenkins

yuxinmmmm:

PR #1511, opened 2021-05-13
 [Submission after feedback] Video Demo: Github integration with Jenkins

yuxinmmmm:


yuxinmmmm:
The changes are based on the feedback from PR #1493 

The changes are:
- slow down the speed of the voice
- reduce some unnecessary steps(eg. running the "git status" command)
- add more explanation between steps 

The new video can be found here [GitHub Integration with Jenkins](https://youtu.be/n3Fp5vN-EjE)

PR #1541, opened 2021-05-30
Repeat_ Essay_ Role of DevOps in minimizing downtime on application using continuous deployment

Renugaa:


Renugaa:
Hi @khaes-kth,

We have now uploaded our repeat essay, but it looks like the check is still being done on the earlier file. How do we remove the old file and just have new one. Please let us know.

//Renugaa

khesoem:
Hi @Renugaa 

Easiest solution: unfork, fork again, change the readme and replace the pdf file (with the same file name), close this PR and create a new one.

PR #1572, opened 2022-03-22
executable-tutorial: proposal GH Action optimization

mellonnen:
# Assignment Proposal

## Title

Optimizing GitHub Action pipelines for node.js applications

## Names and KTH ID
  - Arvid Gotthard (gotthard@kth.se)
  - Bastien Faivre (faivre@kth.se)
## Deadline

Task 1

## Category

Executable tutorial

## Description
We intend to create a step-by-step tutorial on how you could potentially optimize a GitHub Actions pipeline in terms of execution time for a `node.js` application. The idea is to start with a *naive* and slow pipeline and iteratively apply optimizations to it and measure the improvement in execution time.

### DIFF
--- a/contributions/executable-tutorial/gotthard-faivre/README.md
+++ b/contributions/executable-tutorial/gotthard-faivre/README.md
@@ -1,10 +1,20 @@
-# Optimizing GitHub Action pipeline for node.js applications
+# Assignment Proposal
 
-## Members
+## Title
 
-- Arvid Gotthard (gotthard@kth.se)
-- Bastien Faivre (faivre@kth.se)
+Optimizing GitHub Action pipelines for node.js applications
 
-## Proposal
+## Names and KTH ID
+  - Arvid Gotthard (gotthard@kth.se)
+  - Bastien Faivre (faivre@kth.se)
+## Deadline
 
+Task 1
+
+## Category
+
+Executable tutorial
+
+## Description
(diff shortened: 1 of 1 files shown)


javierron:
Hello @mellonnen, 
Thanks for the proposal. I cannot find Arvid Gotthard (gotthard@kth.se) in the course's canvas people page. Have you registered successfully for this course? Or maybe you have unregistered?

mellonnen:
Forgot to register! Did it now!

javierron:
Many thanks! I found you now @mellonnen

Regarding the task, optimizing CI pipelines is a very relevant topic. However, the description could be a bit more specific about which optimizations you plan on showing on the tutorial (a short list naming them is enough).

mellonnen:
Here is the list of optimizations we already found. Note that we're always looking for new ones, our researches are not done yet.

- Cache dependencies
    - this is the main optimization, and the one we expect to be most effective.

- Cache build and test caches
    - jest makes use of a cache to improve the speed of tests. If this cache is persistent between runs there is likely to be some improvement
    - yarn has a global cache that it uses to improve installation of dependencies. Yarn also has a build cache to improve the time of builds. The idea is to keep these persistent between pipelines.

- Run independent jobs in parallel
    - a linting job and a test job could be run in parallel as they do not depend on each other

- Abort irrelevant jobs
    - the idea is to configure pipelines to abort their execution if a new commit is made to the same branch. This does not really optimize for time, rather it is to save resources.

- Run in a pipeline in a custom docker container.
    - The idea is to use a more minimal container like Alpine, instead of a Ubuntu machine. The idea is that the start-up time of the pipeline will be faster.

javierron:
Thanks! Indeed very interesting and useful.

PR #1583, opened 2022-03-23
chore: week folders added for demos and presentations

khesoem:
This adds folders per week for presentations and demos.


Deee92:
LGTM!

PR #1592, opened 2022-03-25
Presentation proposal - Facebook blackout

JustFredrik:
# Assignment Proposal

## Title

Why did Facebook go down in 2021 and could it have been prevented with proper testing?

## Members

- Fredrik Svanholm (svanhol@kth.se)
- Linus Below Blomkvist (libl@kth.se)

## Deadline

Task 1

## Category

Presentation

## Proposal

During October of 2021 Facebooks entire network went down during a period of over six hours. This included all Metas services including Facebook, Whatsapp, Instagram and Sign in with Facebook. this was not due to any server is network malfunction but rather due to human error.
The entire network was theoretically fully operational but was disconnected from the outside world. Many of their systems and interal tools that would´ve been used to revert and fix the issue went down in the process so fixing the issue was not an easy task. 

The problem did not occur imdediatly but gradually cascading out through out the network so catching the problem wasn't that simple.

How did this happen and could it have been mitigated by broad network configuration/ distributed systems testing.


cesarsotovalero:
Hi @JustFredrik, 

This is a very good topic for a presentation.
In which week are you planning to present?
You should edit this PR by adding your user names files into a specific week within the folder "presentation."

PR #1598, opened 2022-03-25
Essay proposal - CI comparison GitHub vs GitLab

nikolim:
# Assignment Proposal

## Title

Continuous Integration Comparison: GitLab vs GitHub

## Names and KTH ID
  - Nikolai Tobias Limbrunner (ntli@kth.se)

## Deadline

Task 1

## Category

Essay

## Description

My goal is to compare the Continuous Integration functionalities of the two largest source control platforms. Initial research has shown that GitLab and GitHub choose different approaches for their Continuous Integration functionalities in some aspects. The aim is to work out the respective weaknesses and strengths of the platforms to draw a short conclusion at the end, for which tasks which platform is better. To create a practical comparison, I plan to create the same CI/CD pipeline using Github and Gitlab respectively, and then also evaluate the process.

### DIFF
--- a/contributions/essay/ntli/README.md
+++ b/contributions/essay/ntli/README.md
@@ -1,0 +1,22 @@
+# Assignment Proposal
+
+## Title
+
+Continuous Integration Comparison: GitLab vs GitHub
+
+## Names and KTH ID
+
+- Nikolai Tobias Limbrunner (ntli@kth.se)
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @nikolim 

Seems to be a good idea. Looking forward to your essay.

PR #1605, opened 2022-03-27
Course automation proposal : Automatic deletion of teammate request

corentinguilloteau:
# Assignment Proposal

## Title

Automatic deletion of teammate request

## Names and KTH ID

-   Corentin Guilloteau (corgui@kth.se)

## Deadline

Task 2

## Category

Course Automation

## Description

This course uses a Github repository issue to allow students to find a teammate. The problem here is that is requires
the students to update or remove their comment on this issue when they have found a teammate. The aim here is to
automate this task.

The automation should perform the following :

-   Run for each new pull request having a label corresponding to an assignment
-   Check if the emails of the author of the task are included in a comment on the teammate finding issue
-   If so, add edit the first comment found by appending `AUTO EDIT: Teammate found`and mentioning the identified users
    to notify them, and comment the PR with a reference to the comment


javierron:
Hello @corentinguilloteau, thanks for the proposal.

The automation action sounds interesting and relevant. I have 2 questions:

- How would you correlate the identities? Consider that not always the person that makes the Pull Request is the same as the one who posts in the "Teammate finder" issue.
- A person could post more than once on the "Teammate finder" issue, in order to find teammates for different tasks. How would you know which comment to remove given that the task category is not always mentioned?

corentinguilloteau:
- To correlate the identities, my idea is to compare the email address specified in each comment on the "Teammate finder" issue with the emails in the pull request.
- Here my idea was to simply remove the first comment found, assuming that the teammate requests would be made in the same order as the tasks.

javierron:
> To correlate the identities, my idea is to compare the email address specified in each comment on the "Teammate finder" issue with the emails in the pull request.

I understand. There is no guarantee that the email will always be present in the comment, however, the action could ignore any comments w/o email.

> Here my idea was to simply remove the first comment found, assuming that the teammate requests would be made in the same order as the tasks.

This is a big assumption. We need to be 100% sure that the comment we remove is the correct one. An incorrect deleting action is extremely inconvenient. I suggest you to explore ideas other than deleting the comments: maybe sending notifications or posting comments where relevant.

javierron:
Also, there is a merge conflict which has to be fixed before merging. Thanks @corentinguilloteau!

corentinguilloteau:
> There is no guarantee that the email will always be present in the comment, however, the action could ignore any comments w/o email.

Indeed, this is a good idea.

> I suggest you to explore ideas other than deleting the comments: maybe sending notifications or posting comments where relevant.

Another idea to solve this issue would be to update the comment by appending something like "AUTO EDIT: Teammate found" instead of deleting the comment.

javierron:
> Another idea to solve this issue would be to update the comment by appending something like "AUTO EDIT: Teammate found" instead of deleting the comment.
I like this. 

Also, a more complete tool would also:
- Comment on the PR, with references to the issue comments (in the teammate finder) which possibly need attention.
- Identify and mention (@) the identified GitHub users, to trigger notifications.

I would accept a more complete Action that implements, in addition, both of the previous points.

javierron:
@corentinguilloteau
Are you still going forward with this task? If so, please update the description addressing the previous discussion.

corentinguilloteau:
@javierron 
I updated the proposal according to our discussion.

javierron:
@corentinguilloteau great, looking forward to see your work.

PR #1621, opened 2022-03-29
added slides used in the presentation

ArnPellesGit:
Here's mine (Per Arn) and Philip Salqvist's (@psalqvist) final PR for the presentation we had today. Basically added the slides as a PDF in our folder + added a link to the same slides in the README.


ArnPellesGit:
I am not sure but I don't think that I can do anything about the missing student registration thingie above...

PR #1669, opened 2022-04-04
Presentation Proposal: An overview of continuous training in MLOps with some insights into Kubeflow

vishalned:
# Assignment Proposal

## Title

An overview of continuous training in MLOps with some insights into Kubeflow

## Names and KTH ID

- Vishal Nedungadi (vishaln@kth.se)

## Deadline

Deadline for task 1

## Category

Presentation

## Description

CD in MLOps is a very interesting and common topic. We found that MLOps differs from Devops because it contains an extra step called Continuous Training. We felt this might be a good topic to provide some insights on. We also delve into Kubeflow and see how that would help in CI/CD/CT for MLOps


vishalned:
Not so sure about what the failed check is. Could someone help us please. Thanks

PR #1670, opened 2022-04-04
Presentation: Continuous Training in MLOps with Kubeflow

vishalned:
# Assignment Proposal

## Title

An overview of continuous training in MLOps with some insights into Kubeflow

## Names and KTH ID

- Vishal Nedungadi (vishaln@kth.se)

## Deadline

Deadline for task 1

## Category

Presentation

## Description

CD in MLOps is a very interesting and common topic. We found that MLOps differs from Devops because it contains an extra step called Continuous Training. We felt this might be a good topic to provide some insights on. We also delve into Kubeflow and see how that would help in CI/CD/CT for MLOps


javierron:
Hello @vishalned, regarding the questions you had during the lab session:
- There is a chance for you to present on week 4, given that there are not enough presentations/demos regarding that week's topic. Otherwise you can present on week 9 since the topic fits there as well. 
- It is possible for you to get an A, since the deadlines for presentations and demos are each week, and are separate from deadlines 1-5. So you could still fulfill deadlines 2-5 + the presentation, and thus get 5 tasks approved.

vishalned:
Hey @javierron thanks for your response on my questions. Its great that we can still get an A. We can present this on Week 9, as it fits well with the topic of that week. Thanks

PR #1736, opened 2022-04-07
Presentation proposal: Integration Testing with Testcontainer

b06902047:
# Assignment Proposal

## Title

Integration Testing with Testcontainer

## Names and KTH ID
  - Mayuri Salunke (salunke@kth.se)
  - Yen Chen (yenc@kth.se)

## Deadline

Task 2

## Category

Presentation

## Description

Nowadays, applications are often composed of micro services that communicate with each other and lots of integration points have to be tested. Comparing with unit test, integration test is often much harder and more time-consuming for developers.  However, we found out that Testcontainer is a useful tool for providing lightweight, throwaway instances of external components that run in dockers. 

We are going to start with briefing about integration testing and how it usually be done. We will then explain why using a container is better, followed by introducing Testcontainer, a useful tool for integration testing on containers. This is an opportunity for us to show how Testcontainer is beneficial in development.

### DIFF
--- a/contributions/presentation/week4-containers-and-serverless/salunke-yenc/README.md
+++ b/contributions/presentation/week4-containers-and-serverless/salunke-yenc/README.md
@@ -1,0 +1,14 @@
+# Assignment Proposal
+## Title
+Integration Testing with Testcontainer
+## Names and KTH ID
+ - Mayuri Salunke (salunke@kth.se)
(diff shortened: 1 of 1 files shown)


javierron:
Hello @b06902047, the proposal looks good to me.

Your presentation will be scheduled with other container-related topics on week 4.

PR #1778, opened 2022-04-14
proposal essay

yani122:
# Assignment Proposal

## Title

Security in containers

## Names and KTH ID

 - Rafi Youssef (rafiy@kth.se)

## Deadline

Task 3

## Category

Essay

## Description

Containers are importent in software development community because they allow developers to avoid the time consuming configuration. The security of conatiner is the process of implementing tools and policies to ensure that container infrastructure and other are protected.

The essay aims to cover what is security related challenges when using containers in DevOps and Security in Cloud Computing.

### DIFF
--- a/contributions/essay/rafiy/README.md
+++ b/contributions/essay/rafiy/README.md
@@ -1,0 +1,24 @@
+# Assignment Proposal
+
+## Title
+
+Security in containers
+
+## Names and KTH ID
+
+ - Rafi Youssef (rafiy@kth.se)
+ 
+
+## Deadline
+
+Task 3
+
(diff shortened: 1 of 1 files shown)


javierron:
@yani122 thanks for the proposal. Both containers and security are very relevant for the course.

Since you target "security in containers" in general, it's expected that the work is not focused only on Docker, but to go over other container-related tools as well, it could be for deployment (Docker, lxd, containerd, etc), orchestration (docker-swarm, docker-compose, kubernetes, etc). 

The essay must be self contained, so remember to provide an introduction and background about containers and security.

PR #1800, opened 2022-04-18
Final Submission - Tutorial: Optimizing your Docker Images

thomassrour:
# Assignment Proposal

## Title

Tutorial: Optimizing your Docker Images

## Names and KTH ID
  - Thomas Srour (srour@kth.se)
## Deadline

Task 2

## Category

Executable Tutorial

## Description
As Docker images can easily get very big very fast, for this task, I would like to make a Katacoda tutorial describing the many different ways of optimizing/ reducing the size of Docker images, for example minimizing the number of layers, caching and multistage builds. 

Final Submission Links:

Katacoda-Scenario Repository: https://github.com/thomassrour/katacoda-scenarios

Katacoda-Scenario Tutorial: https://www.katacoda.com/kyrieswerving

What I tried to achieve:

|                                                                                                                       | Yes                              | No  |
| --------------------------------------------------------------------------------------------------------------------- | -------------------------------- | --- |
| executable: The tutorial can be automatically executed from beginning to the end, in the browser or in CI (see below) | **Mandatory** :heavy_check_mark: | -   |
| ilo: The tutorial states the intended learning outcomes.                                                              | **Mandatory** :heavy_check_mark: | -   |
| motivation: The tutorial is clearly motivated (why it matters for Devops?)                                            | **Yes** :heavy_check_mark:       | No  |
| browser-based: The tutorial can be successful executed in the browser (katacoda is recommended)                       | **Yes** :heavy_check_mark:       | No  |
| ci-based: The tutorial can successful be executed as a CI job                                                         | **Yes**        | No  |
| background: The tutorial gives enough background                                                                      | **Yes** :heavy_check_mark:       | No  |
| illustrated: The tutorial is illustrated with an informative figure (eg a flowchart)                                  | **Yes**                           | No  |
| pedagogical: The tutorial is easy to follow                                                                           | **Yes** :heavy_check_mark:       | No  |
| original: The tutorial is original, no or few similar tutorials exist on the web                                      | **Yes** :heavy_check_mark:       | No  |
| easter-eggs: The tutorial contains an easter egg                                                                      | **Yes** :heavy_check_mark:       | No  |
| language: The language is appropriate (structure, grammar, spelling)                                                  | **Yes** :heavy_check_mark:       | No  |


mrbgco:
@Deee92 @thomassrour 
## Intro
I went through the executed tutorial, and generally speaking, it's a beginner level and requires basic knowledge of Docker even though it briefly explained what even is Docker image/container.
The tutorial proceeds with step 1, which showcases how to reduce image size, layers, and why that is important for business.
After step 2, it clarifies the importance of multi-stage building and how it does add value in easily smooth steps.
Lastly, it approaches caching topics, and how it helps in speeding up building Docker images.

Below, I  categorize my findings:

### *Strengths:*
1- Well organized tutorial, starts with basic Docker optimizations, followed by multi-stage building then a more advanced topic which is caching.
2- Language is easy to understand, and technical terms are explained well and briefly.
3- It covers the most used low-hanging fruits of Docker optimization which is a nice take-home lesson for you.
4- Within each guided step, everything is clear, the building up, and then the conclusion.

### *Weaknesses:*

#### *Major:*
1- There was mentioned that the tutorial would show how the number of layers gets better, but only compare sizes of images were done between the two images of the first step.
**Action:**
So if expanding here to show that tweaking run statements didn't only reduce the size, but the number of layers also. Using "docker history".

2- In step 3, it wasn't clear that the user should change the filename in Dockerfiles while pasting, so as a suggestion:
**Action:**
- Use a fixed name, and keep it simple.
- Highlight that the user should change the filename.

#### *Nitpicking:*

3- It would be easier in step 3, to provide image-building commands.
**Action:** 
Add the missing commands.


4- In the tutorial path, use different names for Docker images to avoid overriding them, so you can conclude at end of the tutorial what are lessons learned with images names.
**Action:** 
Change Docker images names for different steps.

5- There is an option in Katacoda to execute commands in cmd directly:
That will make even the tutorial easier to go through and faster. Check (https://www.katacoda.community/control-sequences.html).
**Action:** 
Add execute for all commands.

6- If expanding on different options of small images then maybe something like distroless is worth mentioning in text.
**Action:** 
Expand on provided info, and mention other options.

7- Worth mentioning also, that using multistage buildings, and small images is a good aspect of Security.
**Action:** 
Expand on the importance of tutorial and multi-stage building security-wise and why it adds value, by stripping useless packages and dependencies.

8- For 1st step, use a smaller image like Alpine to make it even faster. As the showcase is the RUN/ADD/COPY.
**Action:** 
Switch base images of step 1. 

9- No added benefit or running containers of the Docker images in Steps 2, 3. The focus here is on comparing the size of the final image.
**Action:** 
Remove the steps or showcase why it's a needed step.
 
 10- No added value from adding value inside text files in step 3.
**Action:** 
Remove the steps or showcase why it's a needed step.
 
 11- Images in the last step are a bit big, fewer layers could be used to just show that caching worked.
**Action:** 
Use just one package.

PR #1813, opened 2022-04-19
Essay submission: Configuration languages - A comparison of Dhall and CUE

LukasGutenberg:


bpalagi:
Hi, I'd like to do a feedback on your essay with @dchahuan, may we have your permission?

LukasGutenberg:
works for me, @olofeldre ?

olofeldre:
Go ahead!

bpalagi:
# Feedback for Configuration languages essay - Dhall and CUE

## Summary
	
It was a good and informative essay that showed the comparison between two newer configuration languages. It gives a nice criteria on how to judge this new approach. It also gives some good snippets on what can be done. But sometimes when the actual configuration is shown it needs further explanation so the reader can fully understand the intent behind that specific configuration.

## High Level Strength and Weaknesses

### Strengths

<ul>
	<li>The essay comes to a pretty clear conclusion on a preferred configuration language given certain circumstances</li>
<li>It gives clear direction for the purpose of the essay </li>
	<li>The structure allows the reader to easily read and follow the author's intentions</li>
<li>The use of in text references is beneficial to a reader wondering where certain information was found</li>
</ul>


### Weaknesses 

<ul>
	<li> When showing code that the reader may not understand, it should give a high level explanation of what it is that is doing
</li>
<li>The topic of the essay is a comparison so the essay should focus on how the both languages differ from each other.
</li>
	<li>Lack of interactivity with links and in text references inside pdf</li>
</ul>

## Actions to Take

While it is necessary to thoroughly explain the languages, it would be beneficial to the reader to further expand on the comparison because it is the main topic of the. Specifically, expand on the paragraph which directly compares the “Desired Properties” of each configuration language. And show why you would prefer one over the other in this specific case. You did a great job on the feature that was missing in Dhall but missed to explain the difference in the approach for the rest of the configuration file. 

When drawing the conclusion you can use more explicitly the criteria you show on the essay. This can give a more consistent looking criteria and would also make it clearer to the reader why and how you arrived at the stated conclusion. 

The references are nice in that they allow the user to further research a topic, although they are not easily accessible directly through their links on the PDF. Improving both those links and possibly even linking the in text references could make it easier for a reader to research further.
	
Even though YAML engineering is not a formal term, referencing a not reviewed nor trusted source as reddit humor may not be the best way to give academic credibility to the essay and may look like you made unsubstantiated claims.



## Additional Pointers

[PyCue](https://github.com/tebeka/pycue)
[dagger.io](https://dagger.io/)
[Dhall Kubernetes](https://github.com/dhall-lang/dhall-kubernetes)

<ul>
	<li>We would like to see an example of how to actually implement this new configuration languages in an actual project of our own. So we can get and feel the benefits they provide in comparison to the currently dominant configuration languages. Here is a library that allows you to import CUELang configuration with Python 
</li>

<li>
	We were thinking if an actual tool is already implementing this new configuration language in a production environment and found Dagger. A CI/CD tool that uses CUE as the default configuration language and actually lists the same benefits from it as you do.  
</li>
	<li> We found that Dhall also has a github repo to support Kubernetes. It may be worth mentioning, as currently a reader with interest in K8s may decide on CUE because they are unaware of Dhall support. 
</li>
</ul>

PR #927, opened 2021-03-23
Presentation on monitoring with the tool Prometheus

arxra:
No decided topic, but we promise to make it interesting!

PR #1960, opened 2022-05-13
Presentation proposal: From metrics to insight: Prometheus monitoring system overview

gogo432754:
# Assignment Proposal

## Title

From metrics to insight: Prometheus monitoring system overview

## Names and KTH ID

  - geomal@kth.se 

## Deadline

Deadline task 4

## Category

Presentation

## Description

With DevOps becoming more and more complex monitoring becomes much more difficult to handle manually. New system architectures such as micro-services can consist of containerized infrastructures with many processes. Many of the systems that we use are interconnected. The more complex the systems become the more important it becomes to have some form of insight into where errors could occur och where more resources are needed. These problems can be solved with Prometheus, a powerful and open-source platform that makes monitoring a system much easier. I will attempt to explain why monitoring is important and give an overview of why Prometheus is a good tool to use.

### DIFF
--- a/contributions/presentation/week9-other/geomal/README.md
+++ b/contributions/presentation/week9-other/geomal/README.md
@@ -1,0 +1,21 @@
+# Assignment Proposal
+
+## Title
+
+From metrics to insight: Prometheus monitoring system overview
+
+## Names and KTH ID
+
+  - geomal@kth.se 
+
+## Deadline
+
+Deadline task 4
+
+## Category
+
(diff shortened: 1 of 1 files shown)


Deee92:
Hi @gogo432754, we've had a similar presentation before (see #927).
Please ensure that in addition to the overview, you also focus on specific capabilities of Prometheus, and how it can be configured and integrated into an application.

PR #1971, opened 2022-05-16
Final Submission - Essay: Kubeflow or MLflow - which one to choose?

Akseluhr:
# Assignment Proposal

## Title

Kubeflow or MLflow - which one to choose?

## Names and KTH ID
  - Aksel Uhr (auhr@kth.se)
  - Nikos Smyrnioudis (nsmy@kth.se)

## Deadline

Deadline task 4

## Category

Essay

## Description

We would like to start describing and motivating MLOps on a general level, and afterwards narrowing it down to two different and popular MLOps frameworks which we will compare. The frameworks we are thinking about are Kubeflow and MLflow. MLOps is a recent field and as of today, there is no standarized way of doing it. For this vision, we want to clarify, highlight and conclude on the differences between the two mentioned frameworks. We hope that this essay will provide directions for the readers to choose a framework that suits their project interests.  

Due to word constraints, we were not able to include Seldon Core in our comparison. I hope this is fine.


javierron:
Thanks for the submission!

mrbgco:
@Akseluhr Kindly change the filename, it causes problems on Windows due to filename length exceeding the default limit.

PR #2009, opened 2022-05-18
Essay Proposal: What is the Docker Container Lifecycle?

pdchao23:
# Assignment Proposal

## Title

What is the Docker Container Lifecycle?

## Names and KTH ID
  - Preston Chao (pdchao@kth.se)

## Deadline

Deadline task 5

## Category

Essay

## Description

To understand Docker containers, it is vital to understand how the container lifecycle works and the different stages a container can be in. In this essay I will walk through the container lifecycle explaining each step separately detailing how a container falls into the state and the attributes attached to each stage. I will also note the common misconceptions like the difference between pause and stop, and rm and kill.


khesoem:
@pdchao23 : Thanks for the proposal. It is relevant, but lacks novelty. Can you make it a bit more focused and concrete? For example, have a look at these previous proposals: https://github.com/KTH/devops-course/pulls?q=is%3Apr+container+lifecycle+is%3Aclosed

PR #2064, opened 2023-03-20
emilk2-pierrelf: Demo proposal, week 4

saffronjam:
# Assignment Proposal

## Title
Switching to Podman from Docker in your CI/CD pipeline

## Names and KTH ID
  - Pierre Le Fèvre (pierrelf@kth.se)
  - Emil Karlsson (emilk2@kth.se)

## Deadline
- Week 4

## Category
- Demo

## Description
Like Docker, Podman is a containerization tool that allows developers to create, run, and manage containers. Unlike Docker it can, among other things, run as non-root out of the box since it does not need a running Docker daemon. 

This demo will show how a developer could switch to use Podman instead of Docker and take advantage of the benefits that brings with it.

### Relevance
Podman is a tool that manages containers and services, which are the foundation of these modern application development paradigms.

### DIFF
--- a/contributions/emilk2-pierrelf/demo/week3/README.md
+++ b/contributions/emilk2-pierrelf/demo/week3/README.md
@@ -1,0 +1,22 @@
+# Assignment Proposal
+
+## Title
+Switching to Podman from Docker in your CI/CD pipeline
+
+## Names and KTH ID
+  - Pierre Le Fèvre (pierrelf@kth.se)
+  - Emil Karlsson (emilk2@kth.se)
+
+## Deadline
+- Week 4
+
+## Category
+- Demo
+
+## Description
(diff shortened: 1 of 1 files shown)


khesoem:
@saffronjam & @pierrelefevre
You cannot two tasks within the same group. I'm closing this PR as you two have another proposal together for presentation as well.

pierrelefevre:
@khaes-kth "Group work is encouraged (max 2 persons) but you cannot be with the same person for **more than 2 projects**. You can work alone for one or at most two projects."

Should be ok for 2 projects?

khesoem:
You're right! I thought the limit was just one project. Please recreate a PR adhering to new rules in the proposal PR template.

PR #2068, opened 2023-03-21
jonassav-ponorman: Demo proposal, week 2

jonassavas:
# Assignment Proposal

## Title

_CI-pipeline demonstration with GitHub Actions_

## Names and KTH ID
  - Pontus Norman (ponorman@kth.se)
  - Jonas Sävås (jonassav@kth.se)

## Deadline

_Week 2_

## Category

_Demo_

## Description

_The demo will serve as a way for everyone to get familiar with the fundamentals of using Continous Integration with automatic building and testing of a simple Java application using the build tool Gradle with GitHub Actions. We will talk about the requirements of automatic build tools and demonstrate one configuration by adding dependencies and creating the workflow yml-files to build and test the simple application on different operating systems upon a Push or Pull Request to the default branch. The use of the GitHub Actions results will be shown in a 'dummy' Pull Request where our test suite now indicates that the added feature is ready to be merged without breaking the application._

### Relevance (Not included in the README)

_An important part of DevOps is testing and making sure that the desired build is working for every platform that we want it deployed on._
_Thats why we think it would be relevent to demonstrate the CI fundamentals and configure GitHub Actions to automatically build a simple program for different operating systems that also require the installing a few dependecies._

### DIFF
--- a/contributions/demo/week2/jonassav-ponorman/README.md
+++ b/contributions/demo/week2/jonassav-ponorman/README.md
@@ -6,6 +6,7 @@
 _CI-pipeline demonstration with GitHub Actions_
 
 ## Names and KTH ID
+
   - Pontus Norman (ponorman@kth.se)
   - Jonas Sävås (jonassav@kth.se)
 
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @Jonassavas 

Thanks a lot for submitting your proposal. The topic is relevant, but it seems to me that it is not specific enough. It is doing something very general and there are many many similar demos on Youtube for example. Please make a new PR with a more specific topic.

PR #2083, opened 2023-03-22
tmatts-gakk week 2 presentation proposal

RubberHux:
# Assignment Proposal

## Title

Arcmutate: Improving code with mutation testing for Java

## Names and KTH ID

- Tobias Mattsson (tmatts@kth.se)
- Georgios Akkogiounoglou (gakk@kth.se)

## Deadline

- Week 2

## Category

- Presentation

## Description

Traditional code coverage, such as line coverage, is often not enough. This is because it is only capable of checking which code is executed, but it is not able to detect faults. For this reason, things like mutation testing have become relevant. Mutation testing is a software testing technique that involves small changes to a program's source code to evaluate the effectiveness of the testing state in detecting those changes. One modern tool which is focused on providing mutation testing for code bases is *Arcmutate*, which builds and improves upon the open-source project Pitest. Arcmutate introduces a series of mutations or small changes to the original code, and then runs the test state against the mutated code. If the test state detects the mutation, then it is considered "killed" and the test suite is considered effective. If the mutation goes undetected, then it is considered "lived" and the test state is considered not enough. Arcmutate offers integration for a number of things, such as Azure DevOps, Bitbucket, GitHub and GitLabs. 

**Relevance**

Since Java is an undeniably popular language in the industry, it goes without saying that a tool meant for Java mutation testing would be relevant. Arcmutate works for both Gradle and Maven, which are two of the most popular build tools for Java. All in all, the tool offers fast and effective mutation testing for code bases of any size, making it a good tool for Java projects.

### DIFF
--- a/contributions/presentation/week2/tmatts-gakk/README.md
+++ b/contributions/presentation/week2/tmatts-gakk/README.md
@@ -6,16 +6,16 @@ Arcmutate: Improving code with mutation testing for Java
 
 ## Names and KTH ID
 
-Tobias Mattsson (tmatts@kth.se)
-Georgios Akkogiounoglou (gakk@kth.se)
+- Tobias Mattsson (tmatts@kth.se)
+- Georgios Akkogiounoglou (gakk@kth.se)
 
 ## Deadline
 
-Week 2
+- Week 2
 
 ## Category
 
-Presentation
+- Presentation
 
 ## Description
 
--- a/contributions/presentation/week2/tmatts-gakk/README.md
+++ b/contributions/presentation/week2/tmatts-gakk/README.md
@@ -1,0 +1,27 @@
+# Assignment Proposal
+
+## Title
+
+Arcmutate: Improving code with mutation testing for Java
+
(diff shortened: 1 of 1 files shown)


Deee92:
Hi @RubberHux. In your presentation, do you plan to give an overview of Arcmutate and compare it with pitest, or also speak about your experiences using it? I see that it's not a free tool.

RubberHux:
Hi @Deee92, our plan is to primarily give an overview of Arcmutate (how it works, what it offers etc.) since we have no experience of our own with it. This does indeed involve comparing it with pitest, since they are closely related. Also, since Arcmutate offers unique things like running mutation tests directly on pull & merge requests through the git integration, we figured it's interesting to talk about even with no personal experience with it.

Deee92:
Okay, that should work. It will also be interesting to highlight specific aspects on how pitest and Arcmutate differ (e.g., extra mutation operators, and code snippets if applicable). Looking forward to it!

PR #2100, opened 2023-03-28
jonassav-ponorman: Demo proposal, week 3

jonassavas:
# Assignment Proposal

## Title

Canary deployment with Kubernetes and Microsoft Azure

## Names and KTH ID

 Pontus Norman (ponorman@kth.se)  
 Jonas Sävås (jonassav@kth.se)

## Deadline

- Week 3

## Category

- Demo.

## Description

We want to demonstrate what a canary deployment could look like. We want to use Microsoft Azure and Kubernetes to do this.
During the demonstration, we make changes to an app that we'd created and deploy this changed build as a canary deployment.
We then do some tasks on this app and compare some metrics with the previous version and check if the canary version is ready for deployment.

**Relevance**  
Functionality of a deployed application is highly relevant to preserve with the DevOps pipeline and further ensuring that new features of an application works as intended where Canary deployment is one way to fully test the new version where a failed result only affects a limited number of users.

### DIFF
--- a/contributions/demo/week3/jonassav-ponorman/README.md
+++ b/contributions/demo/week3/jonassav-ponorman/README.md
@@ -15,7 +15,7 @@ Canary deployment with Kubernetes and Microsoft Azure
 
 ## Category
 
-- Demo.
+- Demo
 
 
 ## Description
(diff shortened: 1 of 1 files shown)


Deee92:
Sounds good, merging!

PR #2102, opened 2023-03-28
adammel-gardaa Presentation week 3: Rolling Deployment with Kubernetes

Adrelix:
# Assignment Proposal

## Title
Rolling Deployment with Kubernetes

## Names and KTH ID
Adam Melander (adammel@kth.se)
Gard Aasness (gardaa@kth.se)

## Deadline
Week 3

## Category
Presentation

## Description

Rolling deployment is a deployment strategy which is used to ensure low downtime and a smooth update for a application. This is done by slowly replacing previous version of the application with a newer one, one-by-one.

Kubernetes is a container orchestration platform used to automate the deployment of applications among other things. It is a cluster management system, making it easy to replace deployed nodes with new ones running the new version.

Rolling deployment, as with other deployment strategies, is an important deployment strategy that decreases risk and downtime for deployment. This is highly relevant when maintaining any project with low downtime tolerance.


javierron:
Hello @Adrelix 
This is a relevant topic for the course. Blue/Green deployment has been covered before on the course.
However, we have not seen a demo with Kubernetes.  Could you consider changing the type of task?

Adrelix:
Hi @javierron 
We already have an approved demo proposal that we would prefer to do, however we think that it could still be a lot of new content in a presentation about the combination of Blue/Green deployment in combination with Kubernetes rather than just strictly Blue/Green Deployment, would that be ok?

javierron:
@Adrelix 
OK, I understand. What about discussing a less popular/well known deployment strategy, then?

Adrelix:
@javierron Sure, would it be possible to discuss something like Canary or Rolling strategy instead?

javierron:
@Adrelix Rolling deployments have not been discussed previously on the course.
You can change the proposal, and I will merge it.

Adrelix:
@javierron Fixed!

PR #2108, opened 2023-03-30
oliverle-praneetk Presentation Proposal week 3

lehtihet:
# Assignment Proposal

## Title

Orbs make CircleCI the superior CI/CD option

## Names and KTH ID

- Oliver Schwalbe Lehtihet (oliverle@kth.se)
- Praneet Kala (praneetk@kth.se)

## Deadline

- Week 3

## Category

- Presentation

## Description

We want to compare CircleCI against many other state of the art CI/CD tools, such as Jenkins, GitLab CI/CD and more. We want to make the case that CircleCI is better from an ease of use perspective by comparing the Orbs (reusable build tools) in CircleCI to the equivalent tools in Jenkins and other services. We will showcase an example of an Orb relating to Kubernetes or some other Orb to show how easy it is to use. 

**Relevance**

Tools for creating CI/CD jobs are very relevant to DevOps since they automate a lot of the operation processes and integrate it with the developers. It's important to select the best CI/CD tool.

### DIFF
--- a/contributions/presentation/week3/oliverle-praneetk/README.md
+++ b/contributions/presentation/week3/oliverle-praneetk/README.md
@@ -1,0 +1,27 @@
+# Assignment Proposal
+
+## Title
+
+Orbs make CircleCI the superior CI/CD option
+
+## Names and KTH ID
+
+- Oliver Schwalbe Lehtihet (oliverle@kth.se)
+- Praneet Kala (praneetk@kth.se)
+
+## Deadline
+
+- Week 3
+
+## Category
+
+- Presentation
+
+## Description
+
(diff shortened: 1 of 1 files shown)


javierron:
Hello @lehtihet,
The topic is novel and relevant for the course, have you considered making it a demo instead?

lehtihet:
@javierron 
I already have a demo, so no.

javierron:
@lehtihet That's fine. The topic is good. Looking forward to the presentation.

PraneetKala13:
I also have a demo that I will propose for week 4 on containters. Would be great to see a demo on this though.

PR #456, opened 2020-03-30
Presentation week 5: Deployment Strategies

SoFoDa:
#### Members
* David Johansson (davidjo2@kth.se)
* Johan von Hacht (johvh@kth.se)

#### Topic
Comparing two different deployment strategies in the form of Blue/Green deployment and Canary deployment. We find the topic interesting considering that deployment is a key part of DevOps and these two strategies are used within the industry.

#### Initial presentation outline
* Introduction
* Describing Green/Blue deployment
  * How it works  
* Describing Canary deployment
  * How it works
* Comparing the two different strategies
  * Pros/cons
* Take home message

PR #2113, opened 2023-03-30
emihar-mircoram presentation proposal week 3

OmidMircoram:
# Assignment Proposal

## Title

Canary releases vs Feature flag vs Blue-Green deployment. What's the difference and how do you choose between them?

## Names and KTH ID

  - Emil Härdling (emihar@kth.se)
  - Omid Mircoram (mircoram@kth.se)

## Deadline

- Week 3

## Category

- Presentation

## Description

Canary releases, feature flags and blue green deployment are similar deployment methods and can be at first glance seem interchangeable. We would like to discuss their differences and how they might be used in different fields such as social media and the game industry. We also argue how these different methods can affect the userbase and how they can be used to get tailor to specific use cases for the release.

**Relevance**

It's easy to learn new concept in theory but how they are applied in the real world is essential for learning how to and more importantly when to use the different methods. Aside from explaining the methods in theory, they application and result will give insight to other student who might be experts from the computer science perspective but lack a broader view of the use cases.

### DIFF
--- a/contributions/presentation/week3/emihar-mircoram/README.md
+++ b/contributions/presentation/week3/emihar-mircoram/README.md
@@ -28,3 +28,4 @@ and how they can be used to get tailor to specific use cases for the release.
 It's easy to learn new concept in theory but how they are applied in the real world is essential for learning how to and more importantly when to use the different
 methods. Aside from explaining the methods in theory, they application and result will give insight to other student who might be experts from the computer science
 perspective but lack a broader view of the use cases. 
+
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @OmidMircoram 

This seems to be a relevant topic. However, I believe it is very similar to this one: https://github.com/KTH/devops-course/pull/456

Your presentation should not be very similar to a presentation from previous years. So, I'll close this one and look forward to your new PRs.

PR #2126, opened 2023-04-03
loar2-amolt Week 4 Demo proposal 

Leysannn:
# Assignment Proposal

## Title

Implementing multiple VMs with Vagrant

## Names and KTH ID

  - Ley-Olivia Avila Rojas (loar2@kth.se)
  - Ana-Maria Olteniceanu (amolt@kth.se)

## Deadline

- Week 4

## Category

- Demo

## Description

For our demonstration, we will show how you can easily setup multiple virtual machines from one host computer. We will also show how you can sync up the VMs setup from one host computer to another. The VMs will be of different kinds of operating systems and will be running different services. 

**Relevance**
The use of multiple VMs in DevOps is quite widespread, and especially in team context different team members require their VMs setup to be synchronized with each other. Traditionally this is quite difficult to do because they are very large files. Vagrant is one way to streamline and optimize the synchronization of setups.

### DIFF
--- a/contributions/demo/week4/loar2-amolt/README.md
+++ b/contributions/demo/week4/loar2-amolt/README.md
@@ -1,8 +1,9 @@
+
 # Assignment Proposal
 
 ## Title
 
-Implementing serverless architecture with AWS
+Implementing multiple VMs with Vagrant
 
 ## Names and KTH ID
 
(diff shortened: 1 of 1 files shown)


Deee92:
Sounds good, merging!

PR #2200, opened 2023-04-18
ponorman - essay proposal task 3

pontusnorman123:
# Assignment Proposal

## Title

_What does good tests look like in Continuous integration?_

## Names and KTH ID

  - Pontus Norman (ponorman@kth.se)

## Deadline

- Task 3

## Category

- Essay.

## Description

_I would like to delve a bit deeper into continuous integration and more specific the testing part.
I want to investigate what good test looks like but also what less good tests look like and explain why the test 
are good or bad and how they could be improved if bad. I would also like to explore some modern tool that has the
potential to generate test and also look if this is a good way to avoid having bad test in your project._

**Relevance**

_Having test in your project is important to avoid potential problems later on in the DevOps workflow 
. Having good tests is even more important so that we can be as sure as possible that when we deploy our project it works as intended.
That's why it's relevant to know what a good test looks like and if there are tools to create good tests._

### DIFF
--- a/contributions/essay/ponorman/README.md
+++ b/contributions/essay/ponorman/README.md
@@ -1,0 +1,30 @@
+# Assignment Proposal
+
+## Title
+
+_What does good tests look like in Continuous integration?_
+
+## Names and KTH ID
+
+  - Pontus Norman (ponorman@kth.se)
+
+## Deadline
+
+- Task 3
+
+## Category
+
+- Essay.
+
+## Description
+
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @pontusnorman123 

Thanks for the proposal. It is indeed relevant. However, the topic is too broad, something like "what is good integration testing". I will close this PR and look forward to your new proposal with a more specific topic.

PR #2211, opened 2023-04-20
Added Presentation Week 7 Proposal (War Games)

august-tengland:
# Assignment Proposal

## Title

Improving your DevSecOps with War Games

## Names and KTH ID

  - August Tengland (tengland@kth.se)
  - Daniel Hartler (dhartler@kth.se)

## Deadline

Week 7

## Category

Presentation

## Description

As DevOps becomes an increasingly adopted practice, providing sufficent security has become a more complex endeviour. DevSecOps is an approach which seeks to integrate security throughout the IT lifecycle. One principle in DevSecOps is not not only pervent breaches but *assume them* as well. One way to imporve in threat detection and response is to engage in *war games*, in which simulated attacks on the system are ochestrated and managed by the DevOps team. We want to introduce the topic of War Games (more specifically, red & blue teams) and motivate their relevance in the field of DevSecOps.  

**Relevance**

DevSecOps can, and should, be seen as a necessary extension of DevOps principles, as there is a requirement for modern systems to be secure. The change in developement processes has caused a need to rethink security and we belive that war games can be a good practice to complement your DevOps workflows.

### DIFF
--- a/contributions/presentation/week7/tengland-dhartler/README.md
+++ b/contributions/presentation/week7/tengland-dhartler/README.md
@@ -1,0 +1,27 @@
+
+# Assignment Proposal
+
+## Title
+
+Improving your DevSecOps with War Games
+
+## Names and KTH ID
+
+  - August Tengland (tengland@kth.se)
+  - Daniel Hartler (dhartler@kth.se)
+
+## Deadline
+
+Week 7
+
+## Category
+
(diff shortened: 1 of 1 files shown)


khesoem:
Hi @august-tengland 

Thanks for the proposal. Looking forward to your presentation....

PR #2272, opened 2023-04-26
Essay Upload

sdi1400258:
# Assignment Proposal

## Title

Ansible for DevOps and Infrastructure Automation

## Names and KTH ID

  - Georgios Akkogiounoglou (gakk@kth.se)

## Deadline

- Task 3

## Category

- Essay

## Description

Ansible is an open-source automation tool that is widely used in the DevOps community. It allows teams to automate various IT tasks, such as configuration management, application deployment, and orchestration.

**Relevance**

 This essay will explore the various aspects of Ansible, including its features, benefits, and use cases, and will discuss how it can be used to support IaC workflows.


gardaa:
Written by Gard Aasness (gardaa@kth.se)

## Disclaimer
"I certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct."

## High level feedback

### Strengths
- Good structure and language: the report has a good structure in terms of first introducing the report, then the necessary knowledge, and lastly diving deeper into the more advanced sections, which makes it intuitive and easy to follow as a reader. It is also well written. 
- Good introduction and overview: the report has a nice introduction and overview of ansible, which provides the reader with the necessary knowledge to understand the rest of the report. This is crucial for the understanding of the report as a whole, especially if the reader has no prior knowledge about Ansible and IaC.
- The topic is interesting as it covers a very important and central part of DevOps, but at the same time focuses on a specific use case of it. 
- The report uses some images to show some examples of using Ansible 

### Weaknesses
- Personally, I believe the report focused a bit too much about explaining what the different components are, such as ansible and IaC, rather than explaining what the report was actually about: the use of these components in context of CTF. However, I do realize that the word count might restrict the paragraphs, but I think the paragraph length should have been prioritized a bit differently. Although it is important with a introduction to the different components, I think this part could have been cut a bit to give more space to the paragraphs about use cases and best practices. 
- Certain paragrahs of the report was a bit repetetive, and could have been dropped or shortened to avoid redundacy and achieve the goal of the previous point. 
- Although the structure of the report mainly was good, it might have been a better idea to introduce IaC directly after the introduction and before the overview of Ansible, as it is the main foundation and what Ansible can be used for, making it a more natural order in my opinion. 

## Detailed feedback of the different sections

### Abstract
The abstract provides a good, consise and accurate summary of what the reader can expect from the essay, which makes it easy for the reader to get an early indication without having to read the entire report. However, seeing how it is not a scientific article and the total word limit might be a bit low, it may not have been necessary to include it in order to save some words for other paragraphs, but rather make it a part of the introduction instead. 

### Introduction
The introduction is well written. It gives a good introduction to what CTF is and the challenges that can occur, and thereafter clearly motives IaC and Ansible as a solution. It also gives the reader some information about what the essay covers, which is why I believe it is good enough to remove the abstract.

### Overview of Ansible
The overview of Ansible gives a very good introduction to it, which is necessary to understand why it is needed and how it can solve the previously mentioned challenge. The section dives deep into the strengths of ansible and the technical parts of how it works, which emphasizes why it is a state of the art tool for IaC. 

### Infrastructure as Code (IaC)
As mentioned in the high level feedback, I think it would be better if this paragraph was placed before the overview of ansible, to even better understand why Ansible should be used. However, it gives a good overview of what it is. I think the section could be a bit more consise to avoid repetition, for example in the second paragraph (about IaC methodology), where the sentences are very similar and almost states the same. In addition, I feel like the text went a bit out of scope of the section, as the section gives an impression of presenting IaC more generally, but then goes specifically into why Ansible is good for CTF in the third paragraph (about Ansible for CTF). Although I like this paragraph, I think it is better suited to be placed tín the overview of ansible, if the two sections swapped positions. 

### Ansible Use Cases in CTF Games
Interesting to read more about actual use cases for Ansible in CTF. However, I think the section was a bit short, as it is a big part of the report. I really like the figure used to show some code of an Ansible playbook to get a better technical overview and how to actually use it, but I would like to see some more textual explanation about what the code does, as it does not make that much sence if you have no prior knowledge. A bit more detailed description of the Ansible playbooks would also be nice. 

### Ansible Best Practices for CTF Games
It is a good section for the reader, as it provides indication of how to properly use it if they were to try themselves. However, I also think this section is a bit short, as it is a big part of the report, and I would like to know more good practices. I think there is a lack of information some places, for example the second paragraph, where roles are mentioned, but it has not been introduced so I do not really understand what is. I recognize that there is a figure with an example of roles, but a textual description would be nice. Finally, I especially like the paragraph where the author suggests other tools to combine it with, such as Jenkins, because it gives the reader a clear direction of what to look more at. 

### Conclusion
The conclusion summarizes the report well, and reminds the reader of the information in the different paragraphs. However, I do think it could be even shorter and more concise, as for example the last paragraph felt a bit repetitive, seeing how it summarized the conclusion.

sdi1400258:
# Disclaimer
"I certify that generative AI, incl. ChatGPT, has not been used to write this essay. Using generative AI without permission is considered academic misconduct."

PR #2281, opened 2023-04-28
erikwac-jaldu: Trivy executable tutorial proposal

Storsorken:
# Assignment Proposal

## Title

Using Trivy for security scanning

## Names and KTH ID

  - Erik Wachtmeister (erikwac@kth.se)
  - Jaldu Suvarchala (jaldu@kth.se)

## Deadline

- Task 3

## Category

- Executable tutorial

## Description

In our tutorial we want to show how to use a tool called Trivy for security scanning. We intend to show how Trivy can be used in some different contexts, such as a public/private repo, docker image etc.. This would showcase some errors one should try to avoid in development. We'll also showcase an integration with GitHub Actions.

**Relevance**

Security is very important for DevOps/DevSecOps. Using Trivy you could reduce the risk of introducing vulnerabilities in your applications. Further, when you want to make use of some open source software you could use Trivy to check if it has any known vulnerabilities.

### DIFF
--- a/contributions/executable-tutorial/erikwac-jaldu/README.md
+++ b/contributions/executable-tutorial/erikwac-jaldu/README.md
@@ -1,0 +1,26 @@
+# Assignment Proposal
+
+## Title
+
+Using Trivy for security scanning
+
+## Names and KTH ID
+
+  - Erik Wachtmeister (erikwac@kth.se)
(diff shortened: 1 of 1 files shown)


javierron:
Hello @Storsorken, the topic of the tutorial is both relevant and new to the course.

PR #952, opened 2021-03-23
Essay Proposal: FaaS, Serverless, Containerization, and DevOps

Internet-Person-IP:
## Members

Taqui - taqui@kth.se
GitHub: [Internet-Person-IP](https://github.com/Internet-Person-IP)

People will be added...I hope
## Proposal

I want to create an essay where I discuss the basics of FaaS, Serverless Technologies and how they can be used to DevOpsify an application. I also possibly want to talk about the potential for a technology such as [Google Cloud Run](https://cloud.google.com/run)

The topics that will be discussed in the essay:

- Basic Introduction to Cloud
- Basic Understanding of Serverless, relevant terminology and how it relates to DevOps
- How Serverless can be leveraged to simplify the production of large-scale application
- How Serverless can be leveraged to DevOpsify an infrastructure
- How leveraging Serverless and Containerization can be used to simplify the production of large-scale application

PR #2293, opened 2023-05-02
Mircoram Essay proposal

OmidMircoram:
# Assignment Proposal

## Title

Serverless vs Containerization, what is the difference and which one is better for you?

## Names and KTH ID

  - Omid Mircoram (mircoram@kth.se)

## Deadline

- Task 3: Monday, May 8 @ 17h Stockholm Time

## Category

- Essay

## Description

In this essay, I will discuss the pros and cons of both serverless architecture and containers as well as compare them. The essay will discuss for different types of programmes and focus on the architectures rather than one specific programme type. 

**Relevance**

Serverless and containers are both often discussed in devops and their actual application is highly relevant. Comparing them and discussing what types of programmes should use them (or use them together) is an interesting topic.


khesoem:
Hi @OmidMircoram ,

This is a relevant topic. However, we have had a similar essay before: https://github.com/KTH/devops-course/pull/952

So, I'm closing this one and looking forward to your next proposals.

PR #2312, opened 2023-05-06
Removed old proposal

maxennnnn:
Hi,

This was our old PR that was merged in the beggining of the course. We have now presented another subject on an earlier week.

Therefore we wish to remore this presentation as we will not be presenting this subject next week.

Thanks!


SandZn:
All right, I will remove your presentation from week 8. 
.DS_Store files should not be committed. Please make sure "Files changed" contains one file.

PR #2327, opened 2023-05-08
eopru: Essay submission task 3

einaroddurpall:
# Assignment Proposal

## Title

The Purpose and Advantages of Adding Scalability Testing to Your CI/CD Pipeline

## Names and KTH ID

  - Einar Oddur Pall Runarsson (eopru@kth.se)

## Deadline

- Task 3

## Category

- Essay

## Description

Scalability testing is performed to check how well an application scales up or down as a reaction to an increase or decrease
in usage. There are multiple aspects that can be tested, for example, response time, network usage, CPU usage, and more.
One of the primary purposes of scalability testing is finding the limit, how far can you push it until it stops and what is the
reason for it stopping.

**Relevance**

Scalability testing gives a development team an idea of how well the system responds to scaling, giving a better overview
and insight into how the application responds to scaling and can perhaps lead to the discovery of weaknesses.
It can be added to the CI/CD pipeline of the project to be a part of the test automation process.
Being aware of how your system handles scaling is important since poor scalability can for example affect the user's 
experience of using your product or make your product use more resources than it needs leading to an increase in budget.


august-tengland:
## Feedback from August Tengland (tengland@kth.se)
_I certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct._

Thank you for the great text, I thoroughly enjoyed reading it!

### High-Level Strengths:

- Concise presentation: a lot of information is conveyed and the text is not wordy
- Engaging: The presentation style is fun and help maintain interest 
- Comprehensive: The topic is explained, motivated, discussed and put into relation with alternatives
- Convincing: The essay sets out to "sell" the need for scalability testing and succeeds in doing so

### High-Level Weaknesses:

- Outline: Essay structure is not outlined in writing, which makes it somewhat difficult to understand the function of each section 
- Fast paced: The essay moves fast and somewhat abruptly between sections, certain terms and concepts could have been given a bit more care (provided with descriptions, examples etc.)
- Over ambitious: The essay arguably tries to cover to much, and would benefit from being a bit more focused (especially section 2.1)
 
### Introduction:
The introduction is intriguing, fun and clearly motivates the subject. The scenario posed feels realistic, even likely, and does a good job in showcasing both a problem and a proposed solution. The tone is informal and arguably to playful for an academic paper, however "formality" was seemingly not an essay criteria so this should not be considered problematic (and rather helps make the introduction interesting).  The introduction could however be improved by providing a more detailed outline of the report, for example by extending the first paragraph.

The following subsections provide background knowledge about the development of testing methodologies and what scalability actually refers to. Section 1.3 is especially concise and a great primer for someone who hadn't understood  scalability before. All sections are useful however their purpose could have been motivated earlier on, this could arguably be done best by moving them to a separate "background" section where the first header (Background) briefly explains what will be brought up. It also feels like section 1.1 and 1.2 assumes that the reader is familiar with DevOps, which might not hold for the "expected knowledge" of the reader, i.e. a master student in computer science (I for one was unsure of DevOps in my first semester of the master). Providing a full DevOps definition would likely be "overkill" but an brief, inline description could go a long way!

Nitpick: Figure 2 should arguably be referenced to in the text since this is conventional (and what is done with all the other figures).

### Scalability Testing:

The second section describes what scalability testing is and how it is generally conducted. Once again, the presentation is nice an concise. Every sentence has value which means that the text manages to include a lot of consideration in regards to the testing.  As a non-expert, I must admit that it was difficult to distinguish between "system scaling" (increasing the resources) and "system load scaling" (increasing the system load). For example, in the first sentence of the section:

_Scalability testing measures how well the system handles scaling up or down,
and its purpose is to provide information about how well it performs to meet
those demand_

I have difficulty understanding if "scaling up" refers to the systems ability to handle increased loads or its capability to increase resources? This issue could be resolved by further emphasizing the difference between resource-scaling and load-scaling, and which of these two (presumably the latter) we are actually measuring in our tests. 

In the end of section 2.1 (the process), a list of existing tools and some figures are presented. These might contain a lot of interesting information but they are covered so briefly that it is arguably difficult to learn much from their inclusion. It would arguably preferable to use this space to tie up the section or further describe the test execution. 

The "advantages and disadvantages" section provides compelling arguments for scalability testing, especially the first example with customer experience. The disadvantages somewhat general and could likely be tied a bit more to the intricacies of scalability tests as opposed to other tests, consider the cons posed in [this link](https://www.educba.com/scalability-testing/) for example. There is, however, a tight constraint on the length of the essay which can motivate a shorted disadvantages section.

The final subsection "Related Alternatives" gives a solid description on the differences between "scalability" testing and other, similar, testing methods. The distinction between methods is subtle however this is not deemed a fault of the text. 

### Conclusion:
The conclusion is short, sweet and helps reinforce the narrative of the report. A strong take-away message would have been nice, especially since the text has held a conversational tone. The reflection is very short and arguably says to little to motivate it being it's own section, but this is also likely a matter of the word-constraint.

PR #2364, opened 2023-05-22
Added presentation PDF

a-kbd:
Added PDF of presentation as suggested by Martin and Benoit


javierron:
Hello @a-kbd, thanks for uploading the PDF.

PR #2371, opened 2024-08-28
week2: demo proposal

hampfh:
# Assignment Proposal

## Title

Enforcing version consistency between identical package dependencies in monorepos

## Names and KTH ID

- Hampus Hallkvist (hallkvi@kth.se)
- Isak Einberg (ieinberg@kth.se)

## Deadline

- Week 2

## Category

- Demo

## Description

We intend to demonstrate how to enforce version consistency for identical package dependencies in a monorepository, highlighting the benefits and detailing the corresponding implementation. In this demonstration, we will set up three simple web-apps using npm, with overlapping dependencies. We will then create a CI script that performs a tree search across the entire monorepository to compare package versions among the projects. If a mismatch is detected, the CI pipeline will fail, prompting the user to align the version with the rest of the repository. We intend to use a "majority rules" approach, meaning that the version used by most projects will be considered as the correct one, and any attempt to introduce a different version will trigger a failure or warning.

**Relevance**

When working with monorepositories, which store multiple projects in a single repository, one has to be cautious with overlapping dependencies between those projects. These overlaps can cause inconsistencies and conflicts in how different parts of the repository use shared libraries. To avoid these problems, it's important to ensure that all projects within the monorepository use the same version of shared dependencies. This approach maintains consistency throughout the codebase, simplifies dependency management, and ensures that there are no conflicting version of the same dependency.

### DIFF
--- a/contributions/demo/week2/hallkvi-ieinberg/README.md
+++ b/contributions/demo/week2/hallkvi-ieinberg/README.md
@@ -1,0 +1,26 @@
+# Assignment Proposal
+
+## Title
+
+Enforcing version consistency between identical package dependencies in monorepos
+
+## Names and KTH ID
+
+- Hampus Hallkvist (hallkvi@kth.se)
+- Isak Einberg (ieinberg@kth.se)
+
+## Deadline
+
+- Week 2
+
+## Category
+
+- Demo
+
+## Description
+
+We intend to demonstrate how to enforce version consistency for identical package dependencies in a monorepository, highlighting the benefits and detailing the corresponding implementation. In this demonstration, we will set up three simple web-apps using npm, with overlapping dependencies. We will then create a CI script that performs a tree search across the entire monorepository to compare package versions among the projects. If a mismatch is detected, the CI pipeline will fail, prompting the user to align the version with the rest of the repository. We intend to use a "majority rules" approach, meaning that the version used by most projects will be considered as the correct one, and any attempt to introduce a different version will trigger a failure or warning.
+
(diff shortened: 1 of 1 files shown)


javierron:
@hampfh Thanks for the proposal. The directory for the contribution should contain your KTH handles, and not the GitHub handles.

Regarding the proposal, I'm not convinced if it is difficult or original enough. I recommend picking just one of the three tools and present it in depth, including advanced features.

einbergisak:
@javierron We have now changed our project proposal, to one which we regard as both more interesting and complex. 
Thanks!

javierron:
@hampfh Thanks for updating the proposal. I believe this is a better direction.

Can you provide a bit of detail about the CI script? What is the stack you will use to write and deploy it?

hampfh:
> @hampfh Thanks for updating the proposal. I believe this is a better direction.
> 
> Can you provide a bit of detail about the CI script? What is the stack you will use to write and deploy it?

Stack: NodeJS (we are specifically targeting package.json files)
Our sample projects will be simple web applications probably built with [vite.js](https://vitejs.dev/). The applications will just be shells for demonstration purposes.

The script will do a tree search throughout the project, searching for package.json files. We are thinking of writing the script in javascript too and trigger it on every pull-request via Github actions.

hampfh:
@javierron I seem to have accidentally closed this, I just noticed and opened the PR again

javierron:
@hampfh Ok, thanks for clarifying. Looking forward to the demo.

PR #2372, opened 2024-08-29
Week 2: Presentation proposal (samkh & atheers)

Samkth123:
# Assignment Proposal

## Title

Property-based testing in Python using Hypothesis

## Names and KTH ID

  - Sam Khosravi (samkh@kth.se)
  - Atheer Salim (atheers@kth.se)

## Deadline
- Week 2

## Category
- Presentation

## Description

We want to explain the topic of property-based testing, 
where the coder defines the properties that the test cases must satisfy, 
which then automatically generates test cases that enforce these properties. 
We will look at how this is done in practice in Python using Hypothesis, 
which is a library for creating unit tests in Python, based on property-based testing.  
Instead of normal example-based testing, which we have seen in school where we manually define input-output pairs, 
property-based testing instead tests a wider range of inputs to make sure that the code keeps true to the properties the coder has defined.

**Relevance**

Usually, example based testing is used when writing unit tests, where the programmer has to come up with various test inputs and define what the expected results are. 
This is tedious and error-prone in the long run, thus one can utilize property-based testing to speed up the processes. 
Property-based testing also explores more inputs and conditions, which makes it easier to uncover edge cases. 
This is relevant to DevOps as it concerns test automation as well as being able to significantly enhance the testing suite.


javierron:
@Samkth123 Thanks for the proposal. We accept this proposal as discussed during the lab

PR #2373, opened 2024-08-29
Week 3: demo proposal

arejula27:
# Assignment Proposal

## Title

Github Actions workflow for deploying to an AWS Lambda function.

## Names and KTH ID

  - Jacopo Maragna (jmaragna@kth.se)
  - Íñigo Aréjula Aísa (inigoaa@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

In this assignment, we are going to create a Github Actions workflow that will deploy a AWS Lambda function. The workflow will be triggered when a new commit is pushed to the repository. We are going to adopt a modular approach, so we will have different files that can be used in different projects. We are going to set up different workflows that will be conditionally triggered  based on the success of the previous one.

What we aim to demonstrate the possibility to vary the deployment setup with a few slight changes in the configuration files.

**Relevance**

Github actions is one of the most common services for CI/CD. In this case, we will use it to deploy a AWS Lambda function. This is a very interesting use case, as AWS Lambda is a very popular service for serverless computing. We are going to automate all the precess since the code is changed until the final deployment is done.

### DIFF
--- a/contributions/demo/week3/jmaragna-inigoaa/README.md
+++ b/contributions/demo/week3/jmaragna-inigoaa/README.md
@@ -1,0 +1,28 @@
+# Assignment Proposal
+
+## Title
+
+Github Actions workflow for deploying to an AWS Lambda function.
+
+## Names and KTH ID
+
+  - Jacopo Maragna (jmaragna@kth.se)
+  - Íñigo Aréjula Aísa (inigoaa@kth.se)
+
+## Deadline
+
+- Week 3
+
+## Category
+
+- Demo
+
+## Description
+
+In this assignment, we are going to create a Github Actions workflow that will deploy a AWS Lambda function. The workflow will be triggered when a new commit is pushed to the repository. We are going to adopt a modular approach, so we will have different files that can be used in different projects. We are going to set up different workflows that will be conditionally triggered  based on the success of the previous one.
+
+What we aim to demonstrate the possibility to vary the deployment setup with a few slight changes in the configuration files.
+
+**Relevance**
+
+Github actions is one of the most common services for CI/CD. In this case, we will use it to deploy a AWS Lambda function. This is a very interesting use case, as AWS Lambda is a very popular service for serverless computing. We are going to automate all the precess since the code is changed until the final deployment is done.


arejula27:
This proposal corresponds to the pair formed by @jackma-00 and me (@arejula27).

algomaster99:
Hi @arejula27 and @jackma-00! Thank you for your proposal! It is indeed very interesting.
> What we aim to demonstrate the possibility to vary the deployment setup with a few slight changes in the configuration files.

Could you also be specific about the changes? I see there is a template on GitHub if you open the Actions tab and try to setup a workflow for your own repository. Maybe you want to tell something additional to the template there? Or some things we need to be careful in order to avoid failing GitHub actions?

image

jackma-00:
Hello @algomaster99, with our demo we won't use the template to deploy to ECS. Instead, we will come up with our own way to deploy a container to a Lambda function. All will be fully automated and powered through GitHub Actions and AWS CloudFormation. Thanks.

algomaster99:
> we will come up with our own way to deploy a container to a Lambda function

That's good! Please list out shortcomings with the current template also. If you agree, I will approve :D

jackma-00:
The current solution doesn't leverage any pre-configured template or actions from GitHub. It is something developed from scratch which requires a bit of IaC and good knowledge of how AWS works when it comes to deploying new stacks. Therefore it is something not necessarily easy. However, during our demonstration will not go through these complex steps in detail instead we will focus more on presenting a CD strategy to our target environment.

algomaster99:
> during our demonstration will not go through these complex steps in detail instead we will focus more on presenting a CD strategy to our target environment.

So you are proposing a minimal template than the one already on GitHub?

arejula27:
I think that there is a misunderstanding, we are not going to deploy on ECS so there are no drawbacks or advantages between the template and our solution. They are two completely different approaches. 

Our idea is to create a **lambda function**, not a ECS container. These are different AWS services, so we take another workflow for doing it. After our research, there is not official action/template for deploying to lambda (AWS) so we have created our strategy. 

:D

algomaster99:
> After our research, there is not official action/template for deploying to lambda (AWS) so we have created our strategy.

Sounds good! Sorry for the misunderstanding. Merging :D

PR #2394, opened 2024-09-02
Week5: Presentation proposal

ludvigch:
# Assignment Proposal

## Title

_Saltstack: Grains, States and Pillars_

## Names and KTH ID

  - Sofia Edvardsson (sofiaedv@kth.se)
  - Ludvig Christensen (ludvigch@kth.se)

## Deadline

- Week 5

## Category

- Presentation

## Description

We aim to describe how grains, states, and pillars in Salt can be implemented to get infrastructure as code. By describing the relation and different uses of these features a better understanding of how Salt works, and why it is an option for DevOps, can be acquired. We will provide examples of how different configurations impact the overall structure of the infrastructure. Finally we will discuss pros and cons of Saltstack compared to other similar automation frameworks.

**Relevance**

Since Saltstack is a technology used for configuration management and IT automation, among other things, it relates to DevOps. Several of its features, which the presentation will describe and delve into, are using the infrastructure as code paradigm.


algomaster99:
Hi! Good proposal! It would be very nice to include pros and cons over other such automation framework out there. If you agree, I will merge.

ludvigch:
Thank you for the feedback! It sounds like a good idea. We added it to the README and description above!

PR #2395, opened 2024-09-02
week 5: demo proposal

randomicecube:
# Assignment Proposal

## Title

_AIAC: An AI Infrastructure-as-code Generator_

## Names and KTH ID

  - Diogo Gaspar (dgaspar@kth.se)
  - Diogo Melita (diogogm@kth.se)

## Deadline

- Week 5

## Category

- Demo

## Description

We will demo [AIAC](https://github.com/gofireflyio/aiac), a tool which makes use of generative AI tools,
generating infrastucture-as-code "templates, configurations, utilities, queries and more via LLM providers
such as OpenAI, Amazon Bedrock and Ollama".
Our idea is, besides first briefly presenting the tool itself and its capabilities, take one
previously-thought-of prompt and try it in front of the class, and then ask one of our colleagues and/or
one of the teachers for another prompt, and reason regarding the obtained results in both cases (whether they
were good, if re-prompting with a few tweaks should fix basic issues, if it hallucinated, etc).

**Relevance**

Our demo will aim to display this tool's (and generative AI tools' more generally) capabilities
of generating serviceable IaC, and how these may help developers toward their goals (taking, of course,
into consideration that these are fallible tools, with no built-in formal validity/correcteness checks).

### DIFF
--- a/contributions/demo/week5/dgaspar-diogogm/README.md
+++ b/contributions/demo/week5/dgaspar-diogogm/README.md
@@ -1,0 +1,34 @@
+# Assignment Proposal
+
+## Title
+
+_AIAC: An AI Infrastructure-as-code Generator_
+
+## Names and KTH ID
+
+  - Diogo Gaspar (dgaspar@kth.se)
+  - Diogo Melita (diogogm@kth.se)
+
+## Deadline
+
+- Week 5
+
+## Category
+
+- Demo
+
+## Description
+
+We will demo [AIAC](https://github.com/gofireflyio/aiac), a tool which makes use of generative AI tools,
+generating infrastucture-as-code "templates, configurations, utilities, queries and more via LLM providers
+such as OpenAI, Amazon Bedrock and Ollama".
+Our idea is, besides first briefly presenting the tool itself and its capabilities, take one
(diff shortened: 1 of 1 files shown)


algomaster99:
Hi! The proposal looks good to me! Could you please clarify why you are asking for another prompt to the audience? I am not against it, but I encourage it. I just don't understand the reason of doing so. How will it affect the demo?

Also, a piece of advice for your future open-source contributions. Always create a branch on your fork before submitting a PR in order to avoid merging issues later.

randomicecube:
Hey!
The idea was to make it a bit more interactive with the audience, we thought it'd be a cool moment during the demo to engage with both our colleagues and the teachers. Perhaps asking them for a full-blown new prompt would be too hard for a demo, but for example, displaying 2 or 3 possible prompts, asking people to raise their hands for the one which they would think it'd be best to test out and pick out the one picked by the most people would be fun. What do you think?
Also, thanks for the advice!

algomaster99:
>he idea was to make it a bit more interactive with the audience, we thought it'd be a cool moment during the demo to engage with both our colleagues and the teachers.

Yes, I get that and I also think it is super cool. I thought it was essential part of the demo that you NEED some input from the audience. But now I understand that it is for interaction only.

PR #413, opened 2020-03-24
Presentation proposal week 4: A/B Testing

christinerosquist:
# Presentation Week 4: A/B Testing

**Members of the group:** 

Christine Rosquist (rosquis@kth.se, Github: christinerosquist) and Shapour Jahanshahi (shapour@kth.se, Github: shappe)

**Topic:**
A/B Testing in DevOps. 
We would like to learn more about A/B Testing since it seems like an interesting topic that is relevant in DevOps. 

#### Presentation:
- Background/Introduction to A/B Testing, what it is in general
- Why A/B Testing is relevant in DevOps
- Include a concrete technical example: How Netflix uses A/B Testing 
- Reflections and take-home messages

PR #2397, opened 2024-09-02
Week 4 presentation proposal

leegrash:
# Assignment Proposal

## Title

A/B testing of LLMs to increase user experience on a platform using Optimizely

## Names and KTH ID

  - Gustaf Svensson (gussvens@kth.se)
  - Carl Lavö (lavo@kth.se)

## Deadline

- Week 4

## Category

- Presentation

## Description

We want to show the advantages of refining the user experience of LLMs on a platform by A/B testing using the tool Optimizely. The presentation will show the strengths of using A/B testing to analyze usage of the platform and show metrics that can be gathered through Optimizely to tune the LLM specifically to the platforms needs.

**Relevance**

We think that understanding comparative testing methods is important to being able to actually use ML tools in practice. According to the MLOps wikipedia page, very few users of MLOps-practices actually found any real world benefits and being able to reliably test is key in determining whether or not to use it.

### DIFF
--- a/contributions/presentation/week4/gussvens-lavo/README.md
+++ b/contributions/presentation/week4/gussvens-lavo/README.md
@@ -1,0 +1,27 @@
+# Assignment Proposal
+
+
+## Title
+
+The importance of reliable testing methods
+
+## Names and KTH ID
+
+  - Gustaf Svensson (gussvens@kth.se)
+  - Carl Lavö (lavo@kth.se)
+
+## Deadline
+
+- Week 4
+
+## Category
+
+- Presentation
+
+## Description
+
+We want to do a presentation on A/B testing. Explaining why it can provide great value for product owners and developers alike. If it is okay we would like to lean into the business perspective since we are both Industrial Engineering students.
+
+**Relevance**
+
+We think that understanding comparative testing methods is important to being able to actually use ML tools in practice. According to the MLOps wikipedia page, very few users of MLOps-practices actually found any real world benefits and being able to reliably test is key in determining whether or not to use it.


sofiabobadilla:
Hi @leegrash 
The idea of the proposal has potential, yet I would like you to improve both the description and the proposal focusing on how this will be connected to MLOps.

As it is now, the description makes me think you are working on the wrong week (week 2), while it should be week 4. 

FYI: We have seen A/B testing today on week 2 so consider this when designing the presentation.

leegrash:
Hi! 

I understand. The way we see it, A/B testing could either be used to test different iterations LLMS on a platform with users. This is one way we could focus our presentation. Another way is to focus on the systematical testing advantages of LLMS with incremental changes to variables. What way would you suggest us to focus in order for us not to overlap with other presentations or topic weeks?

sofiabobadilla:
A/B testing for LLMs by itself is not very original. 
But if you talk about A/B testing for LLMs using X tool that can be more interesting.

For example this talk from week 2: [Property-based testing in Python using Hypothesis](https://github.com/KTH/devops-course/tree/2024/contributions/presentation/week2/samkh-atheers)

Also, in 2020 someone presented a similar topic ([link](https://github.com/KTH/devops-course/pull/413)) . Maybe you can get more inspiration for the presentation from it :)

leegrash:
I understand. I updated the topic proposal to be more specific regarding tools and purpose. Is this okay?

sofiabobadilla:
The proposal is clear now.
I'll merge.
@leegrash, FYI: I'll send you a calendar invite for your case
Good luck!

PR #2409, opened 2024-09-05
Week 6: Demo proposal

MateusMarinheiro:
# Assignment Proposal

## Title

Using Bitwarden Secret Manager to centrally store, manage, and deploy secrets at scale.

## Names and KTH ID
  - Marco Campione (campione@kth.se)
  - Mateus Marinheiro (mateusma@kth.se)

## Deadline

- Week 6

## Category

- Demo

## Description

In this demo we will:
  - setup a Bitwarden Secret Manager project
  - add a new user to the organization/project
  - add a new secret to the project
  - integrate the secrets created with Github Actions

In the end, we will give a brief explanation of why someone should use Bitwarden secret manager rather than the secrets in GitHub actions.

**Relevance**

This type of tool is very important when it comes to DevSecOps since it streamlines the process of managing secrets and deploying them at scale. It also provides a good way to integrate said secrets into Github Actions which is very useful when it comes to Continuous Integration.


algomaster99:
I like this proposal a lot since I personally use BitWarden. :D
In the demo, I would also want you to motivate on why would someone use Bitwarden secret manager than the [secrets in GitHub actions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions). Is that okay?

MateusMarinheiro:
Yes, it is okay

algomaster99:
Please add this to proposal and then I will merge.

marcocampione:
We have edited the proposal, check if it's ok now @algomaster99

PR #2436, opened 2024-09-05
Week 3: Demo Proposal (samkh & atheers)

Atheer2104:
Assignment Proposal

## Title

Continous Deployment and Rollback using Railway

## Names and KTH ID

  - Sam Khosravi (samkh@kth.se)
  - Atheer Salim (atheers@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

Our demo will show how one can deploy their github project using Railway, we will show how to set up the connection with Railway so that the project becomes deployed. 
Then we will show how the application redeploys when the code changes (continuous deployment). 
Finally we will show one can perform a rollback to previous deployment using Railway. 
Along the way we will also showcase relevant features that Railway provides that are tied with the demo like showing build logs, and deployment logs,
where one can set environment variables, and monitor resource usage (here, we will only show the CPU and memory usage of the deployment). 

**Relevance**

The proposal is relevant to DevOps as it shows how Railway can be used as tool to perform Continuous Deployment (CD) which is a big part of DevOps. 
Also we are showing rollbacks which is also an important feature to have when using Continous Deployment, and finally we are introducing Railway to everyone in the 
class (Railway has never been showcased before) which can be used to deploy various other things like databases and docker images.

### DIFF
--- a/contributions/demo/week3/samkh-atheers/README.md
+++ b/contributions/demo/week3/samkh-atheers/README.md
@@ -1,0 +1,35 @@
+# Assignment Proposal
+
+## Title
+
+Continous Deployment and Rollback using Railway
+
+## Names and KTH ID
+
+  - Sam Khosravi (samkh@kth.se)
+  - Atheer Salim (atheers@kth.se)
+
+## Deadline
+
+- Week 3
+
+## Category
+
+- Demo
+ 
+## Description
+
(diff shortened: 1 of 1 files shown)


sofiabobadilla:
Good proposal. Looking forward to it :)

algomaster99:
Hi! Really good demo. I have a question about it. Could you please tell me how it compares to GitHub action? GitHub actions can run builds. One can also configure secrets. You can also roll back but checking out to a commit and then pushing the commit. What other problem is it solving?

PR #2463, opened 2024-09-11
Add lecture participation tracker using GitHub Actions (#2459)

### LINKED ISSUE #2459: CI job for live counting active participation
Hi all,

We are back with another idea for an open-source contribution. 

This year we introduced [active participation](https://github.com/KTH/devops-course/issues/2370) as one of the criteria for grades. Now, it's a CI course. We won't be counting who asked question in what lecture, right? Let's write a CI script in this repo and whoever does this will get an open source contribution task. Some checks we need to have:

1. Ignore comments by collaborators.
2. Ensure the comment is made on the right date and time of the lecture.
3. Ensure the comment contains the topic that the question is related to (optional)
4. Ensure the comment contains the answer to that question (optional)

Let's start with these criteria and maybe we evolve based on the different comments we see. :D 
@monperrus @sofiabobadilla @javierron @Deee92 Do you want to add more factors above?

bjornthiberg:
# Description

This PR introduces a script and a GitHub Actions workflow to track lecture participation.

The structure of the script and workflow YAML file is based on the existing workflow for updating the Statistics of Student Registrations. It has a similar structure, can be run locally in the same way, and uses the same dependencies.

How it's supposed to work:

GitHub actions workflow
- The workflow is triggered on new issue comments, and proceeds if the issue title is “Lecture participation tracker 2024”.
- track_participation.py is run

Python script (`track_participation.py`):
- Fetches all comments to the issue from the GitHub API
- If the author is a collaborator, or the comment is outside the commenting window (5 hours from the start of each lecture), the comment is ignored.
- Valid comments are stored in a dictionary, with keys for each student who has commented.
- The script generates a markdown table, similar to the one for tracking tasks, which displays the lectures each student has attended.
- The issue description is updated with this markdown table and a timestamp.

I have tested that (in a copy of the repo):
- the workflow triggers correctly when a new comment is added to the tracker.
- comments made by collaborators are ignored as expected.
- comments made outside of the allowed window are not counted.
- valid comments update the participation tracker issue description.
- timezones work as expected (don't quote me on this)

I have not tested for more than two unique students/commenters, as would be the case in "production".

Updating for newer years should only require changing the lecture dates and start hours in the JSON config and the issue number in the YAML file. 

Closes #2459


.github/workflows/lecture_participation.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+jobs:
+  track-participation:
+    if: github.event.issue.title == 'Lecture participation tracker 2024'
```
algomaster99: Since the issue title is mutable, this could break. I suggest comparing equality with the issue number instead.

.github/workflows/lecture_participation.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
algomaster99:
```suggestion
      - uses: actions/checkout@v4
```

We use [latest versions](https://github.com/marketplace/actions/checkout) which reminds me that we need another issue for integrating [renovate](https://github.com/renovatebot/renovate). :)

.github/workflows/lecture_participation.yml
```diff
@@ -0,0 +1,36 @@
...
+      
+      - name: Setup Python
+        uses: actions/setup-python@v2.2.1
```
algomaster99:
```suggestion
        uses: actions/setup-python@v5.2.0
```

.github/workflows/lecture_participation.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          python-version: '3.x'
```
algomaster99: Pin a specific python version. See the importance of [pinning](https://docs.renovatebot.com/dependency-pinning/).

.github/workflows/lecture_participation.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f ./tools/requirements.txt ]; then pip install -r ./tools/requirements.txt; fi
+
+      - name: Set timezone to Europe/Stockholm
+        run: sudo timedatectl set-timezone Europe/Stockholm
```
algomaster99: Why do you need this here? Your python script fetches the comments which have the time in UTC so python will have libraries to convert times, right?

tools/track_participation.py
```diff
@@ -0,0 +1,224 @@
...
+
+    try:
+        collaborators = {collaborator.login for collaborator in tracking_issue.repository.get_collaborators()}
+    except GithubException as e:
+        logging.error(f"GitHub API error: {str(e)}")
+        sys.exit(1)
+
+    for comment in tracking_issue.get_comments():
+        author = comment.user.login
+        comment_time = comment.created_at.astimezone()
```
algomaster99: Here you could convert it to `Europe/Stockholm` timezone instead in the workflow file, right?

tools/track_participation.py
```diff
@@ -0,0 +1,224 @@
...
+# Lecture details
+LECTURE_DATES_TO_NUMBER = {
+    "2024-09-04": 2, "2024-09-11": 3, "2024-09-18": 4, "2024-09-25": 5,
+    "2024-10-02": 6, "2024-10-09": 7
+}
+
+LECTURE_DATES_TO_START_TIME = {
+    "2024-09-04": 13, "2024-09-11": 13, "2024-09-18": 13, "2024-09-25": 13,
+    "2024-10-02": 13, "2024-10-09": 13
+}
```
algomaster99: This is configuration information. Seggrate this from the main script.

algomaster99 (changes requested):
Hi! Thank you so much for the wonderful PR :heart: Please see the changes that I have suggested.

1. What is the default option in your script? `--printPlain`, `--printMarkdown`, `--publish`? Running it without output any options simply exits the script after 3 seconds with code `0`.
3. Do you agree to support maintenance of this script until the end of the course? :) I ask this because there could be a high chance that I merge this PR and then action breaks so I would need your help to fix it :D 


@monperrus @Deee92 @javierron @sofiabobadilla example:
```
+-------+---------------------+-----------------------------+---------------------+
| Index |     Student Name    | Number of Lectures Attended | Lecture(s) attended |
+-------+---------------------+-----------------------------+---------------------+
|   1   |       RafDevX       |              2              |        L2 L3        |
|   2   |       Glassar       |              1              |          L2         |
|   3   |    diogotcorreia    |              2              |        L2 L3        |
|   4   |   MateusMarinheiro  |              2              |        L2 L3        |
|   5   |     bjornthiberg    |              1              |          L2         |
|   6   |       t-sorger      |              2              |        L2 L3        |
|   7   |      jackma-00      |              2              |        L2 L3        |
|   8   |    randomicecube    |              2              |        L2 L3        |
|   9   |        dvavd        |              2              |        L2 L3        |
|   10  |       Tian0602      |              2              |        L2 L3        |
|   11  |      Atheer2104     |              2              |        L2 L3        |
|   12  |      arejula27      |              2              |        L2 L3        |
|   13  |       d-melita      |              2              |        L2 L3        |
|   14  |        M1l0d        |              2              |        L2 L3        |
|   15  | CoordinatesNotFound |              2              |        L2 L3        |
|   16  |    marcocampione    |              2              |        L2 L3        |
|   17  |     Stellariser     |              2              |        L2 L3        |
|   18  |        NoelMT       |              2              |        L2 L3        |
|   19  |     pablo-freire    |              1              |          L2         |
|   20  |  GustavHenningsson  |              1              |          L2         |
|   21  |     Pesteves2002    |              2              |        L2 L3        |
|   22  |     muhammad1928    |              2              |        L2 L3        |
|   23  |       Uqqasha       |              2              |        L2 L3        |
|   24  |       Flopalot      |              1              |          L2         |
|   25  |      bepp-boop      |              1              |          L3         |
|   26  |   Alexanderliu2002  |              1              |          L3         |
|   27  |     laullaurado     |              1              |          L3         |
|   28  |       Jakebobs      |              1              |          L3         |
|   29  |     einbergisak     |              1              |          L3         |
|   30  |      Samkth123      |              1              |          L3         |
|   31  |        kthfre       |              1              |          L3         |
|   32  |      peremateu      |              1              |          L3         |
|   33  |        HexuL        |              1              |          L3         |
+-------+---------------------+-----------------------------+---------------------+
```

bjornthiberg:
Hi @algomaster99, thanks for the feedback!

I just did a force push to get the PR in line with the task structure, by changing the description and adding a task readme file (did not change the python script or YAML file). 

I also added a short documentation to the /tools/README.md file, didn't see that earlier.

1. No default option, it just does the fetching but prints an empty string. Doesn't make a lot of sense, I'll make it print plain text as default and remove it as an option instead.
2. Of course!

I'll get on your other comments asap.

.github/workflows/lecture_participation.yml
bjornthiberg: I thought about how to implement this. One could do it as in [assignment-statistics.yml](https://github.com/KTH/devops-course/blob/87ecf8925262aab78ed8098b1438abadd0a2ee1e/.github/workflows/assignment-statistics.yml), where it's set as an environmental variable and then checked in the python script. The problem with that is that it will unnecessarily continue the CI runner, set up Python, install dependencies etc., even if it's not the correct issue. And I suppose changing a number in the workflow is not more work than changing a repo secret.

So I'll probably just do

`if: github.event.issue.number == 2370`

.github/workflows/lecture_participation.yml
bjornthiberg: Woops, will fix.

needs to be updated in [assignment-statistics.yml](https://github.com/KTH/devops-course/blob/87ecf8925262aab78ed8098b1438abadd0a2ee1e/.github/workflows/assignment-statistics.yml), along with setup python

.github/workflows/lecture_participation.yml
bjornthiberg: Could definitely be done in the Python script instead of using system time, in turn set in the workflow. It's another remnant from assignment-statistics.yml, but probably cleaner to just do it in the script.

.github/workflows/lecture_participation.yml
bjornthiberg: Addressed in commit a0a9a86

.github/workflows/lecture_participation.yml
bjornthiberg: Addressed in commit ab6b4e5

.github/workflows/lecture_participation.yml
bjornthiberg: Addressed in commit ab6b4e5

.github/workflows/lecture_participation.yml
bjornthiberg: Addressed in commit ab6b4e5

.github/workflows/lecture_participation.yml
bjornthiberg: Addressed in e8d3079

tools/track_participation.py
bjornthiberg: Addressed in e8d3079

tools/track_participation.py
bjornthiberg: One solution in 46064c3. 

LMK if having a separate JSON file is not what you wanted, could be in another config.py file as well.

bjornthiberg: (small fix in 79ef6a4)

bjornthiberg:
@algomaster99

I have addressed all your comments now I think! Let me know if I missed something or anything else needs to be adjusted :)

I have tested the code again, and the CI job [on my test repo](https://github.com/bjornthiberg/devops_course_test/actions/runs/10833973286/job/30062022237).

.github/workflows/lecture_participation.yml
algomaster99: I am okay with this approach.

.github/workflows/lecture_participation.yml
algomaster99: Thanks @bjornthiberg ! You can always use "commit suggestion" in order to approve these changes instead of committing separately.
image

contributions/open-source/bthiberg/README.md
```diff
@@ -0,0 +1,53 @@
+# Assignment Proposal
```
algomaster99: Thanks for the proposal, but could you please make a separate PR about this? We take the task and the actual contribution separately.

tools/README.md
```diff
@@ -58,3 +58,32 @@ This script is used to grade each student according to the number of task comple
 
 ### Usage
 ...
+
+
+## track_participation.py
```
algomaster99, line 67: I almost forgot about suggesting this change, but I am happy you noticed! Could you please also add a heading about the prerequisites that should be updated every year?

tools/track_participation.py
```diff
@@ -0,0 +1,215 @@
...
+from prettytable import PrettyTable
+
+PRINT_IN_MARKDOWN = False
+PUBLISH = False
+
+GITHUB_TOKEN = os.getenv("GITHUB_TOKEN")
+REPO_FULLNAME = os.getenv("REPO_FULLNAME")
+ISSUE_NUMBER = os.getenv("TRACKER_ISSUE_NUMBER")
+
+config = json.load(open("tools/track_participation_config.json"))
```
algomaster99:
```suggestion
config = json.load(open("track_participation_config.json"))
```

I think this should be sufficient?

tools/track_participation.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Filename: track_participation.py
```
algomaster99, line 3: May I ask what these headers are for? Is it required for some plugin that you use? I am curious.

tools/track_participation.py
```diff
@@ -0,0 +1,215 @@
...
+    lecture_date_str = comment_time.strftime("%Y-%m-%d")
+
+    if lecture_date_str not in LECTURE_DATES_TO_START_TIME:
+        # comment not made on a lecture day
+        return False
+
+    lecture_start_hour = LECTURE_DATES_TO_START_TIME[lecture_date_str]
+
+    allowed_period_start = comment_time.replace(hour=lecture_start_hour, minute=0, second=0, microsecond=0)
+    allowed_period_end = allowed_period_start + timedelta(hours=COMMENTING_DURATION_HOURS)
```
algomaster99: Sorry if I being too nitpicky here, but I would request you to have two variables here instead of one `COMMENT_DURATION_HOURS`. One which is the actual duration and the other which is the leeway. This would be my personal opinion in terms of code readability/understandability.

algomaster99 (changes requested):
> did a force push to get the PR in line with the task structure, by changing the description and adding a task readme file

Always create a PR from a different branch. You can avoid rebasing in this way.

> Of course!

Thank you!

I think we are almost there! Just a few more changes.

tools/track_participation.py
algomaster99: I am fine with the `JSON` file.

contributions/open-source/bthiberg/README.md
```diff
@@ -0,0 +1,53 @@
...
+I have tested that (in a copy of the repo):
+- the workflow triggers correctly when a new comment is added to the tracker.
+- comments made by collaborators are ignored as expected.
+- comments made outside of the allowed window are not counted.
+- valid comments update the participation tracker in the issue.
+- timezones work as expected (don't quote me on this)
+
+I have not tested for more than two unique students/commenters, as would be the case in "production".
+
+Updating for newer years should only require changing the lecture dates constant in the Python code and the issue title in the YAML file.
```
algomaster99:
```suggestion
Updating for newer years should only require changing the lecture dates constant in the Python code and the issue number in the YAML file. 
```

algomaster99 (changes requested):

tools/track_participation.py
bjornthiberg: Just something I saw in the other python scripts in the repo. I believe the first line is for encoding, but is generally a remnant from python 2 as 3+ uses utf-8 by default. Second line is just for readability I suppose.

bjornthiberg:
> Always create a PR from a different branch. You can avoid rebasing in this way.
 
Damnit, didn't think about this since it's a fork. Could change now but probably more trouble than it is worth since rebasing will not be very complicated.

> I think we are almost there! Just a few more changes.

On it!

contributions/open-source/bthiberg/README.md
bjornthiberg: That makes sense, I have removed the proposal from this PR in dcdb3fd and submitted a separate one (#2480)

tools/track_participation.py
bjornthiberg: It will work when running locally from the tools folder, but the workflow will not find the file as it runs Python in the repo's root dir. 

[See this failed run in the test repo.](https://github.com/bjornthiberg/devops_course_test/actions/runs/10833877493/job/30061700682)

Could change that in the workflow file, but I thought this was cleaner.

tools/README.md
bjornthiberg: addressed in 71c9c9f

tools/track_participation.py
bjornthiberg: Agreed! addressed in ef44d53

bjornthiberg:
@algomaster99 I've addressed the requested changes. Let me know if there’s anything else needed or if you spot any issues! :)

I suppose my commits should be squashed and rebased onto the current 2024 branch. Lmk if I should do that and force push to this PR, to make merging easier.

tools/track_participation.py
algomaster99: > Could change that in the workflow file, but I thought this was cleaner.

I think otherwise because the this path is only true with respect to the CI job so it should be in workflow file only.

algomaster99 (changes requested):
Final change and I will merge it once you resolve it :)

algomaster99 (approved):

algomaster99:
Thanks for the quick fix! Let's give it a run, shall we? I am looking forward to it.

bjornthiberg:
I squeezed in another quick fix! The timestamp ended up being incorrect with the timezone change, should be fixed now in 5c90115

algomaster99:
> The timestamp ended up being incorrect with the timezone change, should be fixed now in https://github.com/KTH/devops-course/commit/5c90115b36fe587ebc92017071e036e1fc114884

Awesome! I did not notice it. Thanks!

algomaster99:
Are we ready to merge? :smile:

algomaster99:
Merging! Let's hope nothing breaks :D

algomaster99:
Great work! One cool thing I immediately realised was that the table could be sorted by the number of lectures attended. Could you also propose a PR for that? I hope it is not too much effort.

bjornthiberg:
Should be easy! Will do next week

algomaster99:
Thanks!

PR #2485, opened 2024-09-13
Week 4: Demo proposal

laullaurado:
# Assignment Proposal

## Title

Using Comet ML to analyze and compare the performance of ML models

## Names and KTH ID

  - Lauren Llauradó (lllr@kth.se)
  - Pere Mateu Raventós (peremr@kth.se)

## Deadline

- Week 4

## Category

- Demo

## Description

The objective of this demonstration is to show the importance of having a way to test and compare the performance of ML models for our applications. 

To do that, we will create two simple machine learning models with a simple dataset, and see that the performance analysis is not a straightforward task, due to having many metrics and ways to analyze it, and each one having its own libraries and structure. Then, we will use Comet ML, which has all the metrics integrated in one single place. Then we will also show how it makes it easy to use them in different applications.

**Relevance**

Being able to easily analyze the performance of your models is crucial, in order to accelerate the model iterations. In addition, comparing models and being able to reproduce tests in multiple scenarios also lets you be consistent and go faster in your machine learning projects.

### DIFF
--- a/contributions/demo/week4/lllr-peremr/README.md
+++ b/contributions/demo/week4/lllr-peremr/README.md
@@ -1,0 +1,28 @@
+# Assignment Proposal
+
+## Title
+
+Using Comet ML to analyze and compare the performance of ML models
+
+## Names and KTH ID
+
+  - Lauren Llauradó (lllr@kth.se)
+  - Pere Mateu Raventós (peremr@kth.se)
+
+## Deadline
+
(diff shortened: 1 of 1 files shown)


javierron:
@laullaurado Thanks for the proposal. Looking forward to the demo. Make sure that it is scoped within the DevOps domain.

PR #2512, opened 2024-09-19
Week 6 Presentation Proposal

chrissap543:
# Assignment Proposal

## Title

Integrating Organizational Policies Using Open Policy Agent

## Names and KTH ID
  - Christopher Sapinski (sapinski@kth.se)
  - Katsutoshi Amano (amano@kth.se)

## Deadline

- Week 6

## Category

- Presentation

## Description

Open Policy Agent is a tool to write and execute policies for different services. These policies check for compliance and encompass a wide variety of options. 
We will highlight the use of OPA to ensure Terraform plans follow policy, while also mentioning the other places that OPA can and should be used. 

**Relevance**
With the wide variety of tools, it can be a nightmare to try and maintain security for all of them. Instead of writing specific policy code for each different product, OPA unifies the toolset. 
This makes it much easier to maintain. These rules can include security specifications such as who can access data.

### DIFF
--- a/contributions/presentation/week6/sapinski-amano/README.md
+++ b/contributions/presentation/week6/sapinski-amano/README.md
@@ -1,0 +1,27 @@
+# Assignment Proposal
+
+## Title
+
+Integrating Organizational Policies Using Open Policy Agent
+
+## Names and KTH ID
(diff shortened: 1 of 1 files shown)


algomaster99:
Looking forward to it :smile:

PR #2546, opened 2024-09-25
Week 6: Presentation Proposal

Pesteves2002:
# Assignment Proposal

## Title

NixOS: Reproducibility with Flakes and Secrets

## Names and KTH ID

  - Tomás Esteves(tmbpe@kth.se)
  - Wenqi Cao(wenqic@kth.se)

## Deadline

- Week 6

## Category

- Presentation

## Description

NixOS is a Linux distribution that follows a declarative approach. This allows it to be reproducible.
However by itself NixOS does not allow to lock packages versions and store secrets securely.
In this presentation We will present the `nix flake` feature that permits us to have a config that does not depend on the time at which it was created
and some tools such as `sops-nix` and `agenix` that allow us to store safely secrets.

**Relevance**

These features and tools are crucial in DevOps, specially in Dependency Management and DevSecOps.

### DIFF
--- a/contributions/presentation/week6/tmbpe-wenqic/README.md
+++ b/contributions/presentation/week6/tmbpe-wenqic/README.md
@@ -1,0 +1,30 @@
+# Assignment Proposal
+
+
+## Title
+
+NixOS: Reproducibility with Flakes and Secrets
+
+## Names and KTH ID
+
+  - Tomás Esteves(tmbpe@kth.se)
+  - Wenqi Cao(wenqic@kth.se)
+
+## Deadline
+
+- Week 6
+
+## Category
+
+- Presentation
+
+## Description
+
+NixOS is a Linux distribution that follows a declarative approach. This allows it to be reproducible.
+However by itself NixOS does not allow to lock packages versions and store secrets securely.
+In this presentation We will present the `nix flake` feature that permits us to have a config that does not depend on the time at which it was created
+and some tools such as `sops-nix` and `agenix` that allow us to store safely secrets.
+
+**Relevance**
+
+These features and tools are crucial in DevOps, specially in Dependency Management and DevSecOps.


Pesteves2002:
@DavidCWQ

algomaster99:
Very well scoped!

RafDevX:
# Feedback

_This is @diogotcorreia's and @RafDevX's feedback for the presentation "NixOS: Reproducibility with Flakes and Secrets" by @Pesteves2002 and @DavidCWQ._

Our feedback is based on slides/transcript in commit https://github.com/Pesteves2002/devops-nixos-presentation/tree/453951dfa4241c948aef3dbf62f6a3b45272032c.

> Disclaimer: We certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct.

---

We would like to start by thanking Tomás Esteves and Wenqi Cao for letting us review their presentation beforehand, providing us with both the slides and the (rough) transcript of what they plan to say about each slide.

## High-Level Strengths

- **Relevant Topic:** as Nix becomes more and more common, it is paramount to address what may very well be the two most significant gaps in vanilla Nix(OS) configuration schemes: the lack of dependency pinning and the lack of secret management. This presentation clearly identifies these two problems and promptly proposes a solution for each of them (flakes and agenix, respectively), with the bonus that both proposed tools are a perfect fit for a Week 6 presentation focusing on DevSecOps and dependency management.
- **Beginner-Friendly:** the presenters do an excellent job of introducing Nix and NixOS to an audience who has never heard of them, debunking common sources of confusion and highlighting the core pillars of the ecosystem. We believe that this presentation does not have any specific prerequisite knowledge, and any DevOps student should easily understand all explanations provided.
- **Engaging Flow:** the presentation's proposed structure is reasonable and easy to follow, especially given the multitude of examples given (including tool usage instructions), which work much better than purely theoretical descriptions. Additionally, we appreciate the care taken to explain in a simple way the high-level structure of a flake, and we believe that the take-home message is both memorable and appropriate in accordance with the rest of the presentation.

## High-Level Weaknesses

- **Unclear Structure:** the presenters should make it more clear at the beginning of the presentation what topics will be discussed and what flow/route they will take, making it easier for the audience to follow. This can be achieved by having an Outline slide at the beginning and using it to quickly explain the intended direction of the talk.
- **Insufficient Motivation:** some of the examples shown are not clearly motivated, and the presenters should clarify their relevance. For instance, it might make sense to mention that the presentation slides are built using Typst (a LaTeX alternative) and that is why Typst is desirable to include in a flake. In addition, it is not very clear why devshells are useful (e.g., separate dependencies for separate projects).

## Content

In addition to what we mention above, there are a few other specific points that we would like to bring up. We think that in general the presentation flows great and fits very well within the week's topic (dependency management & DevSecOps), with concrete examples that allow for a more hands-on experience and reasonably simple explanations of otherwise quite complex technologies.

Nevertheless, there is still room for improvement. In slide 4, the presenters should (orally) clarify what it means for Nix(OS) to be declarative (i.e., that configurations specify an end-goal rather than the means by which they are achieved). Slide 7 feels like it should have something to the right of the text, and the title should likely read "What *are* Nix flakes?". Slide 8 could have a better example output, and slide 11 only shows one host configuration (`spoon`) despite the transcript making reference to several, in addition to requiring a `flake-utils` input which is not used.

Moreover, for Agenix the problem should be more explicit before the solution is presented, and it should be more clear that the users involved are really just machines: slide 15 should better highlight that decryption is unattended (perhaps with a flowchart?), and the corresponding transcript should reference "hosts" instead of "people". Finally, the presenters should clarify that age secrets (encrypted) are meant to be included in version control, pushed to public git repository platforms, and so Agenix is in abstract good for the community since it makes it easier for one to make their configurations accessible by others without risking leaking confidential information.

## Conclusion

In summary, we believe that this presentation has a lot of potential due to its relevance, simplicity, and how appropriate it is to the target audience. We have identified some clear suggestions for improvement, and we are confident that if the presenters choose to take them into account their presentation will both be enjoyable to all and meet the criteria set forth by the course.

Overall, great presentation -- just remember, with Nix, there's no need to _flake_ under pressure! We're sure you'll do great.

PR #2559, opened 2024-09-27
Week 6: Demo proposal

noizy-sthlm:
# Assignment Proposal

## Title

Kubernetes secrets with Sealed Secrets

## Names and KTH ID

  - Amin Nouiser (anouiser@kth.se)

## Deadline

- Week 6

## Category

- Demo

## Description

Managing Kubernetes secrets can be challenging, especially when the infrastructure is defined in code in a public git repository. By default, Kubernetes secrets are only base64 encoded but not encrypted which is not sufficiently secure. Sealed Secrets is a tool that allows secrets to be encrypted by the developer and remain so until they reach the cluster. 

In this demo, I will begin by demonstrating how secrets can be distributed without Sealed Secrets to illustrate the security risk. I will then introduce Sealed Secrets and demonstrate how it solves this problem and makes the distribution more secure.

**Relevance**

This demo is relevant to DevOps as it addresses secure secret management in Kubernetes which is a key principle in DevSecOps.

### DIFF
--- a/contributions/demo/week6/anouiser/README.md
+++ b/contributions/demo/week6/anouiser/README.md
@@ -1,0 +1,27 @@
+# Assignment Proposal
+
+## Title
+
+Kubernetes secrets with Sealed Secrets
+
+## Names and KTH ID
+
+  - Amin Nouiser (anouiser@kth.se)
+
+## Deadline
+
+- Week 6
+
+## Category
+
+- Demo
+
+## Description
+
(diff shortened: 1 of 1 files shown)


algomaster99:
@noizy-sthlm I like the proposal. But could you please find a partner to perform this demo with? It makes our job easier as TAs otherwise we would be assessing a lot more tasks every week, and there already are a lot.

noizy-sthlm:
@algomaster99  Thanks for the suggestion. I thought about teaming up but will be doing the demo solo this time. I appreciate your time.

algomaster99:
@noizy-sthlm the agenda already has [26 tasks](https://github.com/KTH/devops-course/tree/2024/contributions). If we start taking individual task, we would have around 50 presentations in a week which is not feasible.

Anyway, @Deee92 @sofiabobadilla , the proposal is good and I can merge it. Would we be able to find a presentation slot for @noizy-sthlm on Thursday?

PR #2601, opened 2024-10-04
fix link for scientific paper criteria

t-sorger:
There is a wrong link regarding the grading criteria in the README.


sofiabobadilla:
Thanks :)

PR #2556, opened 2024-09-26
Scientific Paper Proposal - Week 6

NicoleWij:
# Assignment Proposal

## Title

Machine Learning-Based Run-Time DevSecOps: ChatGPT Against Traditional Approach

## Names and KTH ID

  - Simon Hocker (hocker@kth.se)
  - Nicole Wijkman (nwijkman@kth.se)

## Deadline

- Week 6

## Category

- Scientific paper

## Description

We will present a recent scientific paper from this year wherein they perform a comparative study between two different approaches to classifying suspicious server log activities and detecting potential threats. Specifically, it compares the classic, traditional machine learning approach of using Weka API for classification with a new proposed novel method of using ChatGPT for performing runtime log analysis. They explore the mechanics and potential of using ChatGPT in Python where context represents labelled data and the questions themselves contain the log records which are being evaluated. Beyond exploring the possibility of it, they also analyse the viability of the novel method, and examine its potential as well as its drawbacks and limitations, and then compares it with the tried and true Weka API method. 

Our presentation will begin by elaborating on why such research is important for the field of DevSecOps, as well as describing the scenario being analysed in the paper. From there, we will go on to detail the workings of both the new and the traditional approach in detail, followed by their respective strengths and weaknesses. Afterwards, we will end the presentation with an explanation of the paper's conclusions.([Machine Learning-Based Run-Time DevSecOps: ChatGPT Against Traditional Approach](https://ieeexplore.ieee.org/document/10192161))

**Relevance**

Security integration in the DevOps pipeline is crucial for maintaining a secure development environment. This paper addresses this need by introducing AI-driven automated log analysis techniques to detect security threats during runtime. It also compares this approach with traditional machine learning practices, providing valuable insights for enhancing security and efficiency in agile development workflows.

PR #2602, opened 2024-10-05
Add feedback proposal

CasperKristiansson:
# Assignment Proposal

## Title

Feedback on scientific paper *Machine Learning-Based Run-Time DevSecOps: ChatGPT Against Traditional Approach*.

## Names and KTH ID

- Casper Kristiansson (casperkr@kth.se)
- Muhammadsodiq Nematjanov (mnem@kth.se)

## Deadline

- Task 3

## Category

- Feedback

## Description

Submitted feedback to [this scientific paper's presentation](https://github.com/KTH/devops-course/pull/2556).

### DIFF
--- a/contributions/feedback/casperkr-mnem/README.md
+++ b/contributions/feedback/casperkr-mnem/README.md
@@ -1,0 +1,22 @@
+# Assignment Proposal
+
+## Title
+
+Feedback on scientific paper *Machine Learning-Based Run-Time DevSecOps: ChatGPT Against Traditional Approach*.
+
+## Names and KTH ID
+
+- Casper Kristiansson (casperkr@kth.se)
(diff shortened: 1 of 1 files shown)


sofiabobadilla:
Hi, 
this task is giving feedback for a task from last week.
For scientific paper, demo and presentations you are supposed to give this feedback before they actually present to the classes.

Did you do that or you send the feedback afterwards?

Also I can not find the feedback proposal? did you do that before submitting?

monperrus:
@CasperKristiansson I confirm that we have two problems: 
- the feedback must be given before the final version, such that authors can take it into account. This presentation was given last week.
- only when the task proposal is merged means that everything is all right and you can start working on it.  Here, this PR has not been merged.

Because of those two reasons, we close this PR and welcome another one.

PR #2624, opened 2024-10-08
Added feedback

TobbeCarlsson:
## Completion of feedback: 

tocarls and lindefors have now given our feedback which can be seen here: https://github.com/KTH/devops-course/pull/2496


javierron:
@TobbeCarlsson Thanks for the submission

TobbeCarlsson:
Does this mean we passed? I can't see grade reported in canvas

TobbeCarlsson:
@javierron

PR #2633, opened 2024-10-09
fix: exclude TAs and Martin from triggering active particpation

algomaster99:
The workflow was triggered when I commented. See https://github.com/KTH/devops-course/actions/runs/11259532298. @bjornthiberg was there some bug in the script that we overlooked? Anyhow, the changes that I have proposed should fix it.

### DIFF
--- a/.github/workflows/lecture_participation.yml
+++ b/.github/workflows/lecture_participation.yml
@@ -9,7 +9,11 @@ permissions:
 
 jobs:
   track-participation:
-    if: github.event.issue.number == 2370
+    if: >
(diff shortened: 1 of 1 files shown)


bjornthiberg:
I suppose it depends on what is intended behavior. I just made it not include collaborators in the actual table. Anyway, changes lgtm!

algomaster99:
> I suppose it depends on what is intended behavior. 

Intended behaviour is to ignore comments by Martin and TAs of devops course.

> Anyway, changes lgtm!

Sounds good. I will merge. This is good for defensive programming.

PR #2644, opened 2024-10-12
Check files affected by task contribution Pull Requests

### LINKED ISSUE #2583: Break CI if there are more than 1 file changes
Some PRs have more than 1 file change. Example: https://github.com/KTH/devops-course/pull/2552. The CI should break if this is the case.

Conditions:
* File changes should be exactly 1. Not sure if we should allow 0 file changes, but let's be strict.

EDIT: updated the list of conditions.

RafDevX:
Fixes #2583 

This pull request checks whether any contributions match the norm of what is expected of a task proposal/submission:

- exactly one file changed, named `README.md` and in a reasonable location; and
- PR description matches the README file (so that reviewers don't need to check both places)

### DIFF
--- a/.github/workflows/check_task.yml
+++ b/.github/workflows/check_task.yml
@@ -3,6 +3,11 @@ name: Check task
 # Controls when the workflow will run
 on:
   pull_request_target:
+    types:
+      - opened
+      - reopened # good to check canvas again to see if eligibility changed
+      - edited # we need to check the new description against the README
+      - synchronize
     branches:
       - 2024
     paths:
--- a/.github/workflows/check_task.yml
+++ b/.github/workflows/check_task.yml
@@ -1,5 +1,3 @@
-# This is a basic workflow to help you get started with Actions
-
 name: Check task
 
 # Controls when the workflow will run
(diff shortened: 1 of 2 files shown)


.github/workflows/check_task.yml
```diff
@@ -12,8 +10,82 @@ on:
...
+          echo "$PR_BODY" > $BODY_FILE
+          if wdiff "$AFFECTED_FILE" "$BODY_FILE"; then
+            echo "Info: PR description matches affected file"
+          else
+            echo "Error: PR description does not match affected file!"
+            exit 1
+          fi
+
+  check-canvas:
+    name: Check proposal is compatible with previous student task registrations
```
RafDevX, line 92: Unsure whether we should require our new check to pass before running the existing Python script.
```suggestion
  check-canvas:
    name: Check proposal is compatible with previous student task registrations
    needs: check-pr-description
```
For now, we have opted not to specify a dependency so that the jobs can run in parallel and be faster for "good PRs", which should be the majority in volume, even if "bad PRs" may suffer slightly since the last job is not aborted even if another one fails first. In any case, the scale is arguably negligible; all the checks are relatively quick.

algomaster99: I agree here. We can run them independently.

.github/workflows/check_task.yml
```diff
@@ -12,8 +10,82 @@ on:
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Install wdiff
+        # we use wdiff (word-diff) instead of plain diff (line-based) to ensure
+        # that we only enforce *content* to be the same rather than specific
+        # file formatting that would otherwise be ignored when the markdown is
+        # rendered. e.g., wdiff allows the README to break lines at col 80 for
+        # readability while still having the PR description without mid-sentence
+        # breaks (since GitHub descriptions don't support them)
```
algomaster99, line 72: What do you mean here? It is unclear. The diff is based on sequence of words, right? 
```suggestion
        # rendered. the comparison takes place between *sequence of words*.
        # eg. "- task 3\n## Category" would be tokenized as ["-task 3", "###Category"]
```
Writing this suggestion, made me question if removes spaces between "-" and "task 3" and so on.

.github/workflows/check_task.yml
```diff
@@ -12,8 +10,82 @@ on:
...
+            exit 1
+          fi
+
+          echo "affected_file=$FILES" >> "$GITHUB_OUTPUT"
+
+  check-pr-description:
+    name: Check whether Pull Request description matches affected file
+    needs: check-affected-files
+    runs-on: ubuntu-latest
+    steps:
```
algomaster99, line 61: This should trigger on description edit as well.

algomaster99 (commented):
Thank you so much! Sorry for the delay in review. Left some questions/comments.

algomaster99:
@RafDevX @ludvigch could you please have a look at my comments?

RafDevX:
Sorry, it's exam season - I'll take a look ASAP today or tomorrow.

algomaster99:
No problem! Thanks for letting me know!

.github/workflows/check_task.yml
RafDevX: Yes, wdiff checks sequence of words, but with a more liberal interpretation of words: anything separated by whitespace is a word, so in your example `- Task 3` is equivalent to `-    Task 3`, but different from `-Task 3` (since then `-Task` would be the word under comparison). This check is fine because `-Task` isn't valid markdown for a list.

.github/workflows/check_task.yml
RafDevX: Fixed in bd6321d30fb564ffed6b567ef5e98e8ea01846c1

.github/workflows/check_task.yml
ludvigch: To be more clear, in the suggestion, `"- task 3\n## Category"` would be tokenized as `["-", "Task", "3", "##", "Category"]` as far as we understand. 

We chose to use wdiff to both workaround the issue of new lines being different in the README and PR description because of Github's automatic formatting of the width as well as to provide students with more  flexibility in formatting with whitespace. At the same time the content is confirmed to be the same and the rendering is probably identical in both cases.

algomaster99: That's clear thanks!

.github/workflows/check_task.yml
```diff
@@ -1,19 +1,96 @@
-# This is a basic workflow to help you get started with Actions
-
 name: Check task
 
 # Controls when the workflow will run
 on:
   pull_request_target:
+    types:
+      - opened
+      - reopened # good to check canvas again to see if eligibility changed
```
algomaster99, line 8: That's a good idea. I think `check-canvas` will take care of it, right?

algomaster99 (approved):
Looks good to me! Haven't tested it, but should work :)

PR #2665, opened 2024-10-15
Executable tutorial submission

oscols:
# Assignment Proposal

## Title

Using Ansible for Disaster Recovery automation

## Names and KTH ID

- Oscar Olsson (oscols)
- Hugo Malmberg (hugomal)

## Deadline

- Task 3

## Category

- Executable tutorial

## Description
This tutorial will be made in KillerKoda. First there will be an introduction to Configuration Management (CM) and Ansible, as well as what it can be used for. Since CM is often used in the military, the theme of this tutorial will be military weapon systems. More specifically, the user will set up two servers for a missile launch system, one of them being a backup. Then, (totally unexpectedly) a missile launch will be activated and the user most act quickly to abort the launch. To achieve this, Disaster Recovery has to be activated, where the backup server will come to use. 

**Relevance**
CM tools are crucial for DevOps environments since they make it possible to configure the environments via scripts. CM can have great benefits, and is not that costly to implement. Therefore, it's important to have knowledge in how to set up CM tools. And Ansible is one of these tools. The benefits of CM becomes evident due to the fact that it enables Disaster Recovery to be implemented to a project, which can potentially save companies in case emergency strikes. And that is why Disaster Recovery is relevant to CM and DevOps as a whole.

KillerKoda: https://killercoda.com/oscols/scenario/ansible-disaster-recovery
Github: https://github.com/oscols/DD2482-tutorial


phlagooo:
# Assignment Proposal

## Title
Feedback for executable tutorial "Using Ansible for Disaster Recovery automation".

## Names and KTH ID

  - Student Oscar Hellgren (oschel@kth.se)
  - Student Johannes Matsson (jmatsso@kth.se)

## Deadline

Task 3
## Category

Feedback
## Description
We would like to give you guys some feedback! :)

Flug0:
Sure, go ahead!

phlagooo:
# Feedback
by oschel@kth.se & jmatsso@kth.se

*I/We certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct.*

## High-level Strengths

### Good Introduction
The introduction of the tutorial gave a good motivation of why configuration management tools, like Ansible, are important to DevOps. It also contained a flow chart giving an overview of the task ahead.

### Intuitive and Checked
The tutorial was pretty intuitive and we had no major problems following it. It also included checks in the process, which was nice to see as it allowed one to move on without worrying about if past errors would affect the next step.

## High-level Weaknesses

### Intent
While the tutorial was easy to follow, we felt that the explanations of why we were doing the tasks could have been motivated more thoroughly. We think that would have made the information easier to remember as one could relate text to tasks mentally.

###Skimming 
The checks were nice to avoid problems, but they also prevented us from being able to skim through the tutorial later to check out specific steps.

### Checks
We did like the checks as a way to ensure problems were caught, but at one point we managed to make a mistake that the check accepted. This later caused a problem and we had to return and fix it. It was specifically the check on “configure server” relating to “backup_server.yml” which erroneously passed, but we suspect other checks could also be flawed.

## Details

### Setting Up Ansible
The installation was easy, but only included instructions for Linux. It would have been nice to also see how it could be done for Mac or perhaps even Windows. The installation process was quite slow, but we understand that problem might be unavoidable.

### Simulate Server
Easy to follow, but we wondered if it has to be set up via Python and use port 8000, or if it merely is what you choose.

### Configure Server
This step was quite large and felt a bit like a wall of text. We feel like the backup could be another step in the tutorial to make each step feel more approachable. But the content itself was quite clear with good explanations of why we did things, like setting file permissions to certain values. We feel like you could have written out the keybinds for Nano instead of saying “Now, you can write out and exit Nano.”, which would have been a bit more clear. We also feel like using cat instead of going into Nano for the backup was somewhat unintuitive as it is different from what we just did, and we mentally expected to repeat the previous steps as it was a backup on the same slide. As we said earlier, we feel like setting up the backup could be its slide, which would make it more intuitive to use a new method. We accidentally entered nano and pasted the whole command, with cat, and wrote out. The check accepted that as we mentioned earlier, which caused problems later.

### Run playbook
This step was clear.

### Disaster strikes
We liked the reverse psychology here, which we also assume is an easter egg. The text was a little sparse, but we still feel like it works as this slide is a bit more cinematic.

PR #2711, opened 2025-08-28
Add demo proposal

Herdi2:
# Assignment Proposal

## Title

Property-based testing with Quickcheck in Haskell

## Names and KTH ID

  - Herdi Saleh (herdi@kth.se)
  - Vincent Lagerros (vinlag@kth.se)

## Deadline

Week 2

## Category

Demo

## Description
The demonstration will begin with a small showcase of property-based testing, which is used
to automate systemic test writing. We will also implement an eDSL (or some sort of custom ADT)
and make it derive the `Arbitrary`  typeclass to allow it to be used for property-based testing within 
the Quickcheck library.

**Relevance**
Property-based testing is used extensively in automated testing. We will specifically demonstrate
how it can be used for custom data types as well, to make it more useful.

Haskell is of specific interest, since most other implementations of property-based testing can
be tracked back to Haskell, and the QuickCheck library.

### DIFF
--- a/contributions/demo/week2/herdi-vinlag/README.md
+++ b/contributions/demo/week2/herdi-vinlag/README.md
@@ -1,0 +1,32 @@
+# Assignment Proposal
+
+## Title
+
+Property-based testing with Quickcheck in Haskell
+
+## Names and KTH ID
+
+  - Herdi Saleh (herdi@kth.se)
+  - Vincent Lagerros (vinlag@kth.se)
+
+## Deadline
+
+Week 2
+
+## Category
+
+Demo
+
+## Description
+This demonstration will show how property-based testing can be used to both lower time
+and increase precision in testing, as well as automate the systemic test writing. We will
+specifically show what property-based testing is and how it works, and potentially show
+some Haskell specific implementations using the Quickcheck library.
+
+**Relevance**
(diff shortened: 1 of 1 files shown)


frankreyesgarcia:
Hi, 

Sound good, but can you specify more about the tests you want to use?

Herdi2:
> Hi,
> 
> Sound good, but can you specify more about the tests you want to use?

We haven't fully developed the test suite yet, but something like:

```haskell
-- Below are some properties to test over arbitrary data
prop_reverse :: [Int] -> Bool
prop_reverse xs = reverse (reverse xs) == xs

prop_add_commutative :: (Int, Int) -> Bool
prop_add_commutative (a, b) = a + b == b + a

prop_sorted :: [Int] -> Bool
prop_sorted = isSorted . quickSort
```
These are some examples. We also want to try and implement a custom data structure and define it as `Arbitrary`  to be able to use that in quickcheck as well!

frankreyesgarcia:
OK, 
But running the test and CI is not enough. 

Scope the proposal on how the tool can be integrated into the CI, with a clear use case.

ericcornelissen:
Thank you for updating, the proposal looks solid now. Looking forward to your demo!

PR #2703, opened 2025-08-28
Week 3 Demo Proposal: Blue-Green Deployment

anica279p:
# Assignment Proposal

## Title

Blue Green Deployment Demo

## Names and KTH ID

  - Riccardo Cocco (cocco@kth.se)
  - Anica Krüger (anicak@kth.se)

## Deadline
- Week 3

## Category
- Demo

## Description

We would demonstrate a Blue-Green Deployment pipeline implemented with two separate Docker Containers and an Envoy Proxy determining the traffic flow. We will have health checks before rerouting the traffic and implementing GitHub Actions.
The demo will show how an application is deployed to a Green env running in parallel to the Blue env (which serves users)
If health checks are passed, the traffic is shifted from Blue to Green with no downtime.

**Relevance**

The topic is connected to the Week 3 topic of Continuous Deployment / Delivery
Blue-Green is a commonly used deployment strategy, and the demo will demonstrate its application. 
It also highlights the importance of rollback strategies, safe releases, and zero-downtime switching.

PR #2722, opened 2025-08-29
Week 3 presentation proposal, rollback in CD

ericwernstrom:
# Assignment Proposal

## Title

The power of canary releases in CD pipelines: Monitoring and Rollback

## Names and KTH ID

- August Yvdal (yvdal@kth.se)

## Deadline

- Week 3

## Category

- Presentation

## Description

We wish to present the process of automated rollback in CD pipelines with canary releases, focusing on how canary releases can help with validating the stability of new versions in production environments. A canary release is a deployment strategy where a new version is released to only a subset of users, by redirecting their traffic away from the current stable version to a “beta release”. Data like error rates and latency can be collected from the canary release, giving the development team an image of how the version performs under real pressure. The presentation will cover the process of collecting data on the canary, the benefits and downsides of this strategy, and what happens when the canary release is not stable enough. For this, we will bring up the example of such a setup using Prometheus as a monitoring tool, and Kubernetes/Flagger as a method of version rollback.

**Relevance**

Despite CI/CD pipelines often having extensive test suites, it is not guaranteed that said tests are up-to-date and able to catch everything which may go wrong when a change is deployed to production. By using canary releases the production/operations and developer teams get to analyze how new releases work on live users. Automated test suites catch many issues before release but they cannot fully replicate real world conditions. Automated rollback mechanisms, as seen with tools like Prometheus and Flagger in Kubernetes, ensure that if a canary underperforms, the system can quickly return to a stable state without manual intervention. This lines up with principles of DevOps like automation and fast feedback.

### DIFF
--- a/contributions/presentation/week3/ericwer-yvdal/README.md
+++ b/contributions/presentation/week3/ericwer-yvdal/README.md
@@ -2,7 +2,7 @@
 
 ## Title
 
-Automatic Rollback in CI/CD pipelines using Kubernetes Flagger and Prometheus
+The power of canary releases in CD pipelines: Monitoring and Rollback
 
 ## Names and KTH ID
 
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Hey,

Thank you for the proposal. However, the description _feels_ more like a demo of using Kubernetes+Prometheus for CD with rollbacks rather than a presentation. It also seems closely related to #2703.

Could you clarify why you think this is a presentation and not a demo and how it differs from  #2703?

ericcornelissen:
Thank you for updating, the proposal looks good and distinct from previous related proposals :+1:

PR #2741, opened 2025-09-04
Week 3 presentation: GoCD

efransholm:
# Assignment Proposal

## Title

Implementing fan-in/fan-out pipelines in GoCD for continous deployment

## Names and KTH ID

- Elin Fransholm (fransho@kth.se)
- Linda Nycander (lindanyc@kth.se)

## Deadline

- Week 3

## Category

- Presentation

## Description

[GoCD](https://www.gocd.org/) is a free and open-source CI/CD server. This presentation will focus on its continuous deployment capabilities, specifically using fan-in/fan-out pipelines to manage large, complex workflows. In large, complex systems, multiple stages can run simultaneously to reduce build and test time, while other stages have dependencies and only occur when all dependencies succeed.

In addition to this, the presentation will also cover automatic material detection, where GoCD can automatically detect which pipelines need to run based on which files have changed. These pipelines allow code to be automatically deployed to stage/production as soon as all dependencies pass.

**Relevance**

Fan-in/fan-out pipelines are essential in modern DevOps since it allows scaling of continuous deployment across multiple services. They reduce pipeline runtime, they are reliable (because of dependencies) and scalable. Fan-in/fan-out in GoCD helps with automation, dependency management and parallel execution to achieve faster and safer continuous deployment.

### DIFF
--- a/contributions/presentation/week3/fransho-lindanyc/README.md
+++ b/contributions/presentation/week3/fransho-lindanyc/README.md
@@ -1,0 +1,28 @@
+# Assignment Proposal
+
+## Title
+
+Implementing fan-in/fan-out pipelines in GoCD for continous deployment
+
+## Names and KTH ID
+
+- Elin Fransholm (fransho@kth.se)
+- Linda Nycander (lindanyc@kth.se)
+
+## Deadline
+
+- Week 3
+
+## Category
+
+- Presentation
+
+## Description
+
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Hey,

This looks like a good, self-contained, focused topic. As a fairly basic concept, the presentation runs the risk of being too simple so I strongly encourage you to come up with advanced/complex use cases, ideally based on the real world.

Best of luck and looking forward!

PR #2732, opened 2025-09-02
Presentation week 3

fridaina:
# Assignment Proposal

## Title
Comparing Deployment Strageties: Shadow, Canary and Blue-Green

## Names and KTH ID
Edwin So (edwinso@kth.se)
Þórfríður Ina A. (tiar@kth.se)

## Deadline
Week 3

## Category
Presentation

## Description
In this presentation, we will compare three deployment strageties used in Devops, Shadow deployment, Canary deployment and Blue-Green Deployment. We will explain breifly how each stragety works, focusing on their pros and cons, and show scenarioes where each method is most suitable. By presenting these strageties side by side, we aim to highlight the trade-offs between risk, cost, user impact and rollback complexity. The comparison will provide an understanding of when to choose one stragety over another. 

Relevance
Deployment strageties are a crucial part of CD as they determine how new software versions are introduced into production. Comparing these three strageties is relevant since each stragety has their own trade offs and therefore understanding the difference can help select the most effective stragety.

PR #2712, opened 2025-08-28
lannge-sendek presentation proposal

CooperUSA:
# Assignment Proposal

## Title

ABsmartly and its Usage for A/B Testing

## Names and KTH ID

  - Lukas Lannge (lannge@kth.se)
  - Samuel Sendek (sendek@kth.se)

## Deadline

- Week 2

## Category

- Presentation

## Description

The presentation introduces the concept of A/B Testing, explaining how it works and what its purpose is. We then discuss how ABSmartly can be used to impliment A/B Testing, and relatate how the program is useful for DevOps.

**Relevance**

A/B testing enables continuous, user-centered improvement by allowing features to be implemented and tested without requiring redeployment. In DevOps, this aligns with the concept of feature flags, which help integrate new features more efficiently. 
ABSmartly supports implementation across multiple platforms and provides flexibility for various testing scenarios, including feature flags, split testing, and multi-variant experiments. This makes the program relevant for DevOps.

PR #2746, opened 2025-09-05
Presentation Proposal Week 3: Comparing Canary Releases and A/B Testing in Progressive Delivery

HaraldHR:
# Assignment Proposal

## Title

Comparing Canary Releases and A/B Testing in Progressive Delivery

## Names and KTH ID
- Harald Hamrin Reinhed (haraldhr@kth.se)  
- Ludwig Flod (lflod@kth.se)  

## Deadline

- Week 3

## Category

- Presentation

## Description
This presentation will explore **two key techniques in progressive delivery**: canary releases and A/B testing. While both are methods for rolling out software gradually, they serve different purposes and are used in different contexts.  We will skip over explanation of how exactly they work since that will have been covered in previous presentations anyway.

The presentation will cover:  
1. Comparison: highlighting key differences:  
   - Goal: reliability vs experimentation  
   - Metrics: error/performance monitoring vs user behavior metrics  
   - Decision-making: rollback vs optimization/insights  
2. Case Studies & Examples: Presenting scenarios where one might be advantagous over the other.
3. Best Practices & Challenges: risk management, statistical significance, monitoring, and rollout strategies.

The goal is to help the audience understand when to use canary releases vs A/B testing, and how both techniques support safer, more informed software delivery.

**Relevance**
This topic is highly relevant to Continuous Delivery and DevOps, as understanding the differences and complementary roles of canary releases and A/B testing allows software engineers with clear strategies for incremental and evidence-driven feature rollout.

### DIFF
--- a/contributions/presentation/week3/haraldhr-lflod/README.md
+++ b/contributions/presentation/week3/haraldhr-lflod/README.md
@@ -1,33 +1,36 @@
-# Assignment Proposal - Week 3
+# Assignment Proposal
 
 ## Title
-**Comparing Canary Releases and A/B Testing in Progressive Delivery**
+
+Comparing Canary Releases and A/B Testing in Progressive Delivery
 
 ## Names and KTH ID
 - Harald Hamrin Reinhed (haraldhr@kth.se)  
 - Ludwig Flod (lflod@kth.se)  
 
 ## Deadline
-Week 3
+
+- Week 3
 
 ## Category
-Presentation
+
+- Presentation
 
 ## Description
 This presentation will explore **two key techniques in progressive delivery**: canary releases and A/B testing. While both are methods for rolling out software gradually, they serve different purposes and are used in different contexts.  
 
 The presentation will cover:  
-1. **Overview of Progressive Delivery** – why gradual rollouts are safer and more effective than traditional releases.  
-2. **Canary Releases** – releasing new code to a small subset of users or servers, monitoring performance, and rolling back if issues arise.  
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Hey,

This proposal looks too long, covering 6 topics in 6 minutes is very difficult. Moreover, most of the topics are already covered (e.g. #2722, #2712, #2732). Please rewrite your proposal to be more realistic for a 6 minute presentation and be distinct from previous proposals, or reconsider your topic.

HaraldHR:
Thanks for the feedback! We've now update it. The idea is to focus less on describing what they are, since that will already have been covered by other groups. Instead we'd like to focus more on examples of when to use one over the other and best practices. I hope this is different enough from the other presentations! @ericcornelissen

ericcornelissen:
Thank you for the update. This now looks distinct enough, even from #2732 because you focus on 1 other technique and _delivery_ instead of _deployment_. I approve your proposal and look forward to your presentation.

Two minor pieces of feedback:

> We will skip over explanation of how exactly they work since that will have been covered in previous presentations anyway.

I do suggest you cover it shortly, e.g. in the form of a 1-2 line definition, to ensure you and your audience have a common understanding of these terms.

> Case Studies & Examples: Presenting scenarios where one might be advantagous over the other.

Case studies would be preferable because they make it easier to argue that there's a true (or at least real-world) advantage. Whereas with an example you'd need to convince the audience your example is realistic.

PR #2800, opened 2025-09-15
Week 6: Demo - Prevent malicious attacks from LLMs

ujanjan:
# Assignment Proposal

## Title

Prevent malicious attacks from LLMs

## Names and KTH ID

  - Fauzan Helmi Sudaryanto (fhsu@kth.se)
  - Bingjie Zhao (bingjiez@kth.se)

## Deadline

Week 6

## Category

Demo

## Description

It's now becoming easier to integrate LLM capability into a product's offering - but is it easy to keep them secure?

With the rise of LLM integrations, it's also important to safeguard the interactions between the user and the system so malicious attacks through LLMs can be prevented. One recent case study revealed a successful takeover of an AI sales agent to reveal its own private customer data through prompt injection [(source)](https://x.com/mbrg0/status/1953815729947447770).

Related to this, we'd like to demo [Superagent](https://www.superagent.sh) which introduces itself as a _proxy_ between a service and the LLM provider to secure both the service owner and the end user. It offers a practical solution to add security protection to prevent prompt injection and personal data leaks through LLM. We want to see whether Superagent is indeed easy to setup and beneficial as they are advertised.

**Relevance**

We think that DevSecOps could be done at any stage of development and is a _continous concern_. Tools like Superagent can help dev team members to secure their system and users in any stage where LLMs are involved in the process.

### DIFF
--- a/contributions/demo/week6/fhsu-bingjiez/README.md
+++ b/contributions/demo/week6/fhsu-bingjiez/README.md
@@ -1,0 +1,30 @@
+# Assignment Proposal
+
+## Title
+
+Prevent malicious attacks from LLMs
+
+## Names and KTH ID
+
+  - Fauzan Helmi Sudaryanto (fhsu@kth.se)
+  - Bingjie Zhao (bingjiez@kth.se)
+
+## Deadline
+
+Week 6
+
+## Category
+
+Demo
+
+## Description
+
+It's now becoming easier to integrate LLM capability into a product's offering - but is it easy to keep them secure?
+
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Hey,

This proposal looks interesting. Could you elaborate on the relevance, in particular why it is relevant to week 6 (_Dependency Management & DevSecOps_)?

ujanjan:
We saw the concept of [Intrusion detection system (IDS)](https://en.wikipedia.org/wiki/Intrusion_detection_system) from this reference list: https://github.com/KTH/devops-course/issues/18#issue-373349478. 

When reading the approach of how Superagent tries to secure systems, it fits with our understanding of `Anomaly-based intrusion detection systems` -- which we think are relevant to week 6 topic.

ericcornelissen:
Thank you for the clarification. Looking forward to your demo!

PR #2788, opened 2025-09-12
Tutorial for DevSecOps on security scanning

anica279p:
# Assignment Proposal

## Title
Automated Security Scanning in CI/CD Pipelines: Building a DevSecOps Tutorial

## Names and KTH ID
  - Anica Krüger (anicak@kth.se)
  - Shreyas Sawai (sawai@kth.se)

## Deadline
Task 2

## Category
Executable tutorial

## Description
This executable tutorial demonstrates how to implement automated security vulnerability scanning in CI/CD pipelines, creating a complete DevSecOps workflow that prevents vulnerable code from reaching production environments.

The tutorial will guide users through building a CI/CD pipeline that automatically scans code for security vulnerabilities using industry-standard open-source tools: Semgrep (Static Application Security Testing), OWASP Dependency Check (dependency vulnerability scanning), and Grype (container image scanning). Users will learn to configure security gates that fail builds when critical vulnerabilities are detected, implement vulnerability thresholds, and create automated security reporting.

The hands-on tutorial will include:
- Setting up a sample application with intentional security vulnerabilities
- Configuring GitHub Actions workflow with integrated security scanning
- Implementing multiple security scanning tools (Semgrep for SAST, OWASP Dependency Check for dependency scanning, and Grype for container scanning)
- Configuring failure thresholds and security policies
- Demonstrating the complete scan-fail-fix-pass cycle
- Creating security reports and vulnerability management workflows

The tutorial will be delivered through KillerKoda platform.

**Relevance**
This tutorial addresses a critical need in modern DevOps practices where security integration (DevSecOps) has become essential for organizational resilience. 

The security approach demonstrated in this tutorial is fundamental to DevOps methodology, where security checks are integrated early in the development lifecycle rather than being an afterthought. This aligns with core DevOps principles of automation, continuous integration, and early feedback loops.

Modern software supply chain attacks and the increasing complexity of dependencies make automated security scanning not just a best practice, but a necessity for any production system.

PR #2729, opened 2025-09-01
Week 6: Presentation Proposal Shifting Security Left in DevOps

xu-zuo:
# Assignment Proposal

## Title

Shifting Security Left in DevOps: Container Vulnerability Scanning with Trivy

## Names and KTH ID

 - Hanzhi Zhang (hanzhizh@kth.se)
 - Xu Zuo (xzuo@kth.se)

## Deadline
- Week 6

## Category
- Presentation

## Description

We want to explore the importance of shifting security left in DevOps pipelines, focusing on how container vulnerability scanning can prevent critical security incidents.

We will explain why traditional DevOps pipelines often ignore security, how vulnerabilities in base images (such as Log4j or outdated Node.js versions) can slip into production, and how automated scanning addresses this challenge. As a concrete example, we will show how Trivy can be integrated into CI/CD to automatically detect vulnerabilities before deployment.

**Relevance**

The rise of DevSecOps shows that security must be part of the pipeline, not an afterthought. Container vulnerability scanning directly addresses this need by ensuring that the artifacts delivered through continuous deployment are safe and reliable. By combining Trivy as a case study, our presentation highlights how teams can combine speed and security, preventing future incidents while maintaining delivery velocity.

PR #2857, opened 2025-09-24
Securing a vulnerable dockerfile

EdwinAhl:
# Assignment Proposal

## Title

Securing a vulnerable dockerfile

## Names and KTH ID

  - Edwin Ahlstrand (edwinahl@kth.se)
  - Lukas Lannge (lannge@kth.se)

## Deadline
- Task 3

## Category
- Executable tutorial

## Description
This project creates an interactive Killercoda tutorial where students learn DevSecOps practices by analyzing and fixing a deliberately insecure Dockerfile. The tutorial begins by providing a suspicious Dockerfile that contains common security flaws, they are then tasked with finding the flaws using both tools such as trivy and analyzing on their own, lastly they are tasked with fixing the flaws.

**Relevance**
Containers are central to modern DevOps pipelines, but insecure Dockerfiles can introduce vulnerabilities such as docker container escapes like CVE-2025-9074. This project teaches how to identify, understand, and fix common container security issues.


ericcornelissen:
Hey, your proposal looks very similar to both #2729 and #2788 and your using the same tools. Please revise your proposal so that it is more original.

EdwinAhl:
@ericcornelissen What if instead of using trivy we used dockle and laid more focus on how to fix the vulnerable dockerfile? So the tutorial would work something like this:
- Scan the dockerfile using dockle
- Interpret the problems
- For each problem, we teach how to manually fix them
- Verify changes using dockle

ericcornelissen:
@EdwinAhl given that you closed the Pull Request, I don't know if you still want to do something like this, but the direction of using a tool like [dockle](https://github.com/goodwithtech/dockle) is interesting - just make sure it hasn't been done before.

EdwinAhl:
@ericcornelissen Thanks for the feedback! We wanted to go in a different direction. Please checkout #2860 instead. If that is rejected we might revisit this idea with your feedback in mind.

PR #2860, opened 2025-09-25
Storing secrets locally using an OpenBao container

EdwinAhl:
# Assignment Proposal

## Title
Storing secrets locally using OpenBao

## Names and KTH ID

  - Edwin Ahlstrand (edwinahl@kth.se)
  - Lukas Lannge (lannge@kth.se)

## Deadline
- Task 3

## Category
- Executable tutorial

## Description
The tutorial teaches how to setup and store secrets locally in an OpenBao container. The steps taken during the tutorial are: 
1. Access a database using a password stored as plaintext in a python program
2. Scan the python file using detect-secrets (pip module)
3. Setup OpenBao docker container
4. Store secret in OpenBao
5. Update the python program to use the secret from OpenBao instead of plaintext

**Relevance**
In terms of DevSecOps, security must be managed by the team developing the product. By using OpenBao as a local docker container, a team can manage secrets safely and locally.


ericcornelissen:
Nice and interesting proposal, I'm looking forward to giving OpenBao a try :)

emilkanerva:
Hey, could @CooperUSA and I give feedback on this?

MT0DE:
> Hey, could @CooperUSA and I give feedback on this?

Hi, @emilkanerva 
Absolutely, we appreciate any suggestions😁

CooperUSA:
# Feedback on “Storing secrets locally using an OpenBao container”

## AI

## Strengths

- Great use of illustrations, they really help clarify what's being typed. 
    - The OpenBao image, in particular, makes it much clearer how the key is split. 
    - We also appreciated that you illustrated the overview with ASCII art instead of just describing it with words. This approach makes it a lot easier to get a good grasp of what is to come. If we ever create another tutorial in the future, we will definitely try to incorporate this technique ourselves. 
- We really enjoyed that most of the tutorial was automated, it made the experience both fun and engaging, as we could simply just press buttons and learn while watching the work happen.
- We thought it was great that you guys included a recap at the end with key takeaways, even though it was not a requirement in the grading criteria. It nicely rounds out the tutorial and encourages the reader to reflect on what they’ve learned.
- We really liked the minecraft gif on the first page, while it wasn’t Subway Surfers or Family Guy it did help with our Gen-Z attention spans!

## Weaknesses

- The tutorial only had two quite long steps. This makes it harder to follow along, and it might’ve been easier to split it into more but shorter steps. One way could be to split the first page/step into three: setup database, setup python & connect to database, and using detect-secrets. The second page could also have been split into for example: OpenBao solution, Storing the MySQL password in OpenBao, and updating the python program to use OpenBao.
- On the second step we are prompted to manually copy a value. This means the tutorial is not fully automatic. Considering the previous feedback point, splitting the tutorial into more steps would also allow using more background and verification scripts to enable automating more parts of the process, such as extracting the secret and providing it to the following python script automatically.
- There is very little reassurance for the user that all commands have been executed properly beyond just seeing their outputs. To improve this, you could explain more what the expected outputs are for each step. You could also use verification scripts which are executed when the user tries to go to the next step, to ensure that the user has executed all the relevant commands, reassuring the user that they’ve executed all commands correctly, and ensuring for the sake of the tutorial that the user will be able to execute everything on the next pages.
- You are very explicit about not using any of the dev keywords when running the server in production, but you do not go into detail about how this would work. A few lines of text explaining a basic outline of how a production deployment would work could be nice.
- While storing passwords in plaintext is obviously bad, the tutorial does not quite highlight the benefit of using OpenBao over for example having the password be entered manually by the user, like in `openbao.py`, but without openbao. Looking deeper into the exact commands executed, it seems the tokens that were created have a time limit. This could have been highlighted to better convince the user of the benefits of a vault solution like OpenBao.

### Pointers

- KillerCoda’s example of utilising verification and the files they used to set it up:
    - https://killercoda.com/examples/scenario/verification
    - https://github.com/killercoda/scenario-examples/tree/main/verification

- OpenBao’s documentation has a page on the security model used, what is and isn’t considered: https://openbao.org/docs/internals/security/

EdwinAhl:
Thanks for the feedback @CooperUSA @emilkanerva! We have implemented your suggestions and updated our submission.

PR #2935, opened 2026-08-27
Scientific paper proposal: FlakyGuard (ASE 2025), week 2

LeuisKen:
# Assignment Proposal

## Title

FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale (ASE 2025)

## Names and KTH ID

  - Jiaxun Wei (jiaxun@kth.se)
  - Matyas Kozar (kozar@kth.se)

## Deadline

- Week 2

## Category

- Scientific paper

## Description

We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context  to resolve flakiness, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.

**Relevance**

Flaky tests break the assumption that a failing build means the code is broken. When working with flaky tests, developers prefer to rerun jobs rather than investigate the failure, and over time the test suite loses its value. Therefore, fixing them automatically is directly relevant to the subject of test automation and CI.

### DIFF
--- a/contributions/scientific-paper/week2/jiaxun-kozar/README.md
+++ b/contributions/scientific-paper/week2/jiaxun-kozar/README.md
@@ -19,7 +19,7 @@ FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale (ASE 2025)
 
 ## Description
 
-We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.
+We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context to resolve flakiness, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.
 
 **Relevance**
 
--- a/contributions/scientific-paper/week2/jiaxun-kozar/README.md
+++ b/contributions/scientific-paper/week2/jiaxun-kozar/README.md
@@ -1,0 +1,26 @@
+# Assignment Proposal
+
+## Title
+
+FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale (ASE 2025)
+
+## Names and KTH ID
+
+  - Jiaxun Wei (jiaxun@kth.se)
+  - Matyas Kozar (kozar@kth.se)
+
+## Deadline
+
+- Week 2
+
+## Category
+
+- Scientific paper
+
+## Description
+
+We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.
+
+**Relevance**
+
+Flaky tests break the assumption that a failing build means the code is broken. When working with flaky tests, developers prefer to rerun jobs rather than investigate the failure, and over time the test suite loses its value. Therefore, fixing them automatically is directly relevant to the subject of test automation and CI.


contributions/scientific-paper/week2/jiaxun-kozar/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+- Week 2
+
+## Category
+
+- Scientific paper
+
+## Description
+
+We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.
```
algomaster99:
```suggestion
We want to present the ASE 2025 paper ["FlakyGuard: Automatically Fixing Flaky Tests at Industry Scale"](https://doi.org/10.1109/ASE63991.2025.00179) by Li, Behrang, Shi, and Liu. The authors propose a tool called FlakyGuard that automatically repairs flaky tests, and they evaluate it in production at Uber. We plan to focus on how FlakyGuard traverses a call graph to collect just enough context to resolve flakiness, discuss the reported success rate together with how many fixes actually land in production, and contrast the approach with previous work in this field.
```

algomaster99 (changes requested):
This looks very well in scope, and the paper's abstract is quite interesting. Just dropped a minor wording change.

LeuisKen:
@algomaster99 Thank you for the suggestion, updated.

algomaster99:
Closing and reopening PR to retrigger CI.

algomaster99 (approved):
@LeuisKen and Matyas, the proposal looks good and you can start making a presentation on it. I am approving it, but I can't merge due to canvas issues which I will fix soon. After the fix, I am happy to merge your PR.

algomaster99:
@LeuisKen and Matyas, sorry for such noisy activity history, but I think I have fixed the bugs. So I am merging it and it hopefully lands in canvas too.

algomaster99:
It does! You can now officially prepare your presentation.

LeuisKen:
@algomaster99 Thanks!

Kozarsson:
Thank you @algomaster99!

Miao0404:
Hi @LeuisKen and @Kozarsson! Jingze Guo (jingze@kth.se) and I (miaoli@kth.se) would like to give you feedback on your scientific paper presentation. Please let us know if that works for you.

LeuisKen:
> Hi @LeuisKen and @Kozarsson! Jingze Guo ([jingze@kth.se](mailto:jingze@kth.se)) and I ([miaoli@kth.se](mailto:miaoli@kth.se)) would like to give you feedback on your scientific paper presentation. Please let us know if that works for you.

Sure, go ahead! I will send you our draft slide when we feel it's ready.

JingzeGuo:
# Feedback
## **Overall feedback**

Overall, we think the presentation is well structured and easy to follow. The connection to DevOps is clear right from the start with the example of flaky tests blocking CI and slowing down releases (slide 2). After that, the group moves to the context problem (slide 3), then walks through the FlakyGuard approach, the experiments, critical analysis, and finally some reflections (slides 4-13). We think this order works very well because after the first few slides it is already quite clear what problem the paper is trying to solve and why it matters.

## **Timing**

One issue we noticed is the timing. The actual presentation took about 12 minutes, which was quite a bit over the time limit. The background section, such as the explanation of what a flaky test is, could be shortened since it was already covered in class.

In slides 6-8, there is a lot of technical detail, and it took quite a long time to explain. This is the paper's core contribution, so it is worth keeping, but it does not need to be explained node by node.

The later sections on the critical analysis and reflection also contained too much text and were a bit too detailed. It would be better to focus on the most important points and explain those more concisely.

## **A few factual points**

There is one factual claim on slide 9 that we think could be changed. The slide says that 100% of developers accepted the root cause explanation. The paper reports something slightly different: 19 developers answered the survey, and all 19 said that the explanations were useful.

We think it would be more accurate to write something like: "19/19 surveyed developers found the explanations useful." The difference matters here specifically because slide 9 puts this right next to the 48.2% fix-rejection number, and that number is also framed in terms of acceptance, except it is about a completely different thing: whether developers accepted the code fix, not whether they found the explanation useful. 

## **Critical analysis**

The critical analysis slide raises several useful points. Instead of only repeating the 47.6% result from the paper, the group also calculates what happens across the complete pipeline. Out of 1115 attempted flaky tests, 197 fixes were finally accepted and landed, which gives roughly 17.7% end-to-end. We think this is a useful number to show because it gives a somewhat different picture of the practical success rate than the headline number alone.

The discussion about generalizability is also useful. All quantitative evaluation is reported on Uber's Go monorepo, and the slide points out that the paper's own Java port is only said to work, with no numbers reported. This is a good catch, since it shows the paper's claim of generalizing to other languages is not quantitatively demonstrated.

There is one limitation on this slide that we would rephrase. The current slide says that fixes only modify test code, so flakiness caused by production code cannot be repaired. The paper's running example actually has the root cause inside a goroutine in the production code, while the final fix is still done on the test side by adding synchronization. We think a more accurate line would be something like: "Fixes only touch test code, so a production-side root cause can sometimes be worked around with a test-side patch, like the paper's own goroutine example, but not resolved if the fix itself has to change the production code."

## **Reflection**

The reflection slide is a good way to end the presentation because it discusses when the approach is useful, when it might not be useful, and what kinds of teams could benefit from it.

Some of the claims could be softened a little. For example, saying that a 17% success rate would be negligible for a startup, or that manual fixes are faster in a small repository, is difficult to conclude from this paper alone. They are reasonable possibilities, but the paper does not really evaluate them.

A presentation is supposed to show what the paper found, so the audience needs to be able to distinguish the paper's findings from your own reasoning beyond what the paper actually tested.


## **Generative AI**

We certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct.

PR #2951, opened 2026-08-30
Demo proposal : Automating dependency management and versioning using NuGet packages

h4lan:
# Assignment Proposal

## Title

Automating dependency management and versioning using NuGet packages

## Names and KTH ID

  - Halan Ouensanga (halano@kth.se)
  - Nolan Blanc (nmsblanc@kth.se)

## Deadline

- Week 6

## Category

- Demo

## Description

We want to present the NuGet package manager, and how it allows developers to create, share, and import public or private dependencies for .NET applications. Developped by Microsoft as the official package manager for the .NET framework, package updates towards remote galleries can be integrated in CI/CD pipelines to automatically provide new versions when new code is merged. It is especially useful for sharing code between repositories: since development teams often choose to split their code into different repos to separate concerns further, it makes it easier to keep their dependencies up to date.

**Relevance**

When companies have people working in parallel on different parts of an app, and packages are regularly updated with new code, versioning, automatic updates, tests, and conflict checks need to be operational. It can be especially tricky if new pushes use code that is not up to date, or when the same dependency is being worked on in two different places.

### DIFF
--- a/contributions/demo/week6/halano-nmsblanc/README.md
+++ b/contributions/demo/week6/halano-nmsblanc/README.md
@@ -1,0 +1,26 @@
+# Assignment Proposal
+
+## Title
+
+Automating dependency management and versioning using NuGet packages
+
+## Names and KTH ID
+
+  - Halan Ouensanga (halano@kth.se)
+  - Nolan Blanc (nmsblanc@kth.se)
+
+## Deadline
+
+- Week 6
+
+## Category
+
+- Demo
+
+## Description
+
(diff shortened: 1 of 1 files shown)


frankreyesgarcia:
That sounds interesting. Could you clarify whether those dependencies are internal or global?

h4lan:
> That sounds interesting. Could you clarify whether those dependencies are internal or global?

If you're asking if the dependencies are public or private, they can actually be **both**.

That's another interesting part: developers can setup their own local server or use the NuGet gallery (online server) to make packages accessible privately. It also poses the question "_how secure is it?_" because if something happens to your code stored remotely, you don't have as much control (and I'm realizing that my be an interesting thing to mention, if there is time for it).
It also allows sharing packages through a network file share, but I did not work with that (yet).

PR #2962, opened 2026-09-01
Demo proposal: Improving Test Quality with Mutation Testing in a CI Workflow

a-runebou:
# Assignment Proposal

## Title

Improving Test Quality with Mutation Testing in a CI Workflow

## Names and KTH ID

  - Alexander Runebou (alerun@kth.se)
  - Love Lindgren (lolindgr@kth.se)

## Deadline

Week 2

## Category

Demo

## Description

Our demo investigates why conventional code coverage is not always sufficient for evaluating the quality of a test suite. We will construct a small application with tests that achieve high or complete statement coverage while still failing to detect meaningful faults.

We will introduce mutation testing in a CI workflow using `mutmut` in Python, where small artificial faults are introduced into the program and the existing test suite is evaluated based on whether it detects them. Surviving mutations will expose weaknesses that are not visible from code coverage alone. 

We will then improve the test suite using techniques such as boundary-value testing and parameterized testing, and rerun the mutation tests to demonstrate how the test suite can improve without necessarily increasing code coverage.

The CI pipeline will use the results as automated validation mechanisms and enforce a quality gate based on the code coverage, test results, and mutation score. During the live demo, we will change the tests and CI workflow to showcase how these changes affect the validation results.

The demo will also discuss some trade-offs of mutation testing, such as computational cost, when mutation testing is appropriate in a CI pipeline, and how it can be balanced against execution time.

**Relevance**

This demo is directly relevant to the topics of software testing, verification, and continuous integration. Automated tests are a central validation mechanism in CI pipelines, but commonly used metrics such as code coverage only indicate whether the code has been executed, but not necessarily whether the tests can detect incorrect behaviour.

Mutation testing provides a way to evaluate the effectiveness of a test suite by checking whether deliberately introduced changes are detected by the tests. Combining mutation testing wih boundary-value and property-based testing demonstrates how different testing techniques can complement one another.

### DIFF
--- a/contributions/demo/week2/alerun-lolindgr/README.md
+++ b/contributions/demo/week2/alerun-lolindgr/README.md
@@ -1,0 +1,37 @@
+
+# Assignment Proposal
+
+## Title
+
+Improving Test Quality with Mutation Testing in a CI Workflow
+
+## Names and KTH ID
+
+  - Alexander Runebou (alerun@kth.se)
+  - Love Lindgren (lolindgr@kth.se)
+
+## Deadline
+
+Week 2
+
+## Category
+
+Demo
+
+## Description
+
+Our demo investigates why conventional code coverage is not always sufficient for evaluating the quality of a test suite. We will construct a small application with tests that achieve high or complete statement coverage while still failing to detect meaningful faults.
+
+We will introduce mutation testing in a CI workflow, where small artificial faults are introduced into the program and the existing test suite is evaluated based on whether it detects them. Surviving mutations will expose weaknesses that are not visible from code coverage alone.
+
+We will then improve the test suite using techniques such as boundary-value testing and property-based testing, and rerun the mutation tests to demonstrate how the test suite can improve without necessarily increasing code coverage.
+
+The CI pipeline will use the results as automated validation mechanisms and enforce a quality gate based on the code coverage, test results, and mutation score. During the live demo, we will change the tests and CI workflow to showcase how these changes affect the validation results.
+
+The demo will also discuss some trade-offs of mutation testing, such as computational cost, when mutation testing is appropriate in a CI pipeline, and how it can be balanced against execution time.
+
+**Relevance**
+
+This demo is directly relevant to the topics of software testing, verification, and continuous integration. Automated tests are a central validation mechanism in CI pipelines, but commonly used metrics such as code coverage only indicate whether the code has been executed, but not necessarily whether the tests can detect incorrect behaviour.
+
+Mutation testing provides a way to evaluate the effectiveness of a test suite by checking whether deliberately introduced changes are detected by the tests. Combining mutation testing wih boundary-value and property-based testing demonstrates how different testing techniques can complement one another.
--- a/contributions/demo/week2/alerun-lolindgr/README.md
+++ b/contributions/demo/week2/alerun-lolindgr/README.md
@@ -22,9 +22,9 @@ Demo
 
 Our demo investigates why conventional code coverage is not always sufficient for evaluating the quality of a test suite. We will construct a small application with tests that achieve high or complete statement coverage while still failing to detect meaningful faults.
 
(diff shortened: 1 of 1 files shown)


ericcornelissen (changes requested):
Interesting proposal, please add which mutation testing tool you will be using (and if you considered others, maybe mention those too) and for what programming language. Also, reconsider whether you want to use property-based testing because introducing it as well within the time span of a demo might be a bit too much, and resolving the mutants with example-based or parameterized unit tests would be sufficient.

a-runebou:
We updated the proposal according to your feedback

ericcornelissen (approved):
Thank you, looks good 👍

EccirilloM:
As agreed privately on emails. Me (emcir@kth.se) and Riccardo Fragale (fragale@kth.se) will do a feedback on your demo.
Can you confirm the agreement also here to make it public?
@a-runebou

a-runebou:
Yes, we confirm the agreement. That is totally fine with us! @EccirilloM

EccirilloM:
Feedback

Technical Depth: During the presentation, the team demonstrated high control of a comprehensive workflow, applying some automation steps to a small program that calculates an item's price based on its weight. The interacting tools (pytest, mutmut, GitHub Actions), while straightforward to use individually, collectively represent a solid system depth. Each component was effective in addressing their proposed problem through a clear, automated solution.

Liveness: Throughout the presentation, they actively interacted with the code, showing their designed workflow in real time. In response to our questions, they proved their ability to extend the Demo and confidently operate the proposed tools by live-editing both tests and functional application code. In particular, they expanded their testing to include more edge cases rather than limiting themselves to the optimal scenario presented before. Furthermore, they were able to show a deep detail about CI results, which showed that they had experimented previously with the tools they proposed.

System Reasoning: During the Demo, the presenters concisely outlined the principal components of their system. Then, during the Q&A, they detailed their proposed architecture more. However, it would be better in our opinion to expand on their primary tool (mutmut) more extensively during the Demo, to let the audience understand a bit better which role it has in the workflow.

Relevance: The presenters clearly told us how their Demo's content reflects the topics covered in class. Specifically, they remarked the importance of CI practices and demonstrated why the core theme of their proposal is essential to their specific use case. In fact, they identified the importance of the analysis of the actual coverage of tests, showing a technique that is able to automatize this procedure in an efficient manner.

Design Decisions: The presenters clearly justified their design decisions for the Demo, elaborating on the trade-offs that drove their choices. Specifically, when asked about a mutmut CI trigger, they clearly justified their implementation strategy by providing a real-world use case that strongly supported their approach (triggering a lot of Mutation tests might slow down dangerously the pipeline and so additional tests are called only in some commits).

Reflection: The presenters clearly explained the limitations of their Demo and scenarios where their showcased tools are less effective in the testing pipeline of an application. However, they could highlight the potential applications more and strengths of the concepts presented in their Demo.

Narrative: The narrative certainly needs improvement. Adding a couple of introductory slides to clearly define the problem being addressed and the proposed solution would help the audience follow the Demo's storytelling with greater focus. On the other hand, they were quite precise regarding the practical implications of the workflow they presented.

Speech: The discussion and the topics presented were clear and easy to understand. However, the interaction between the two presenters could be improved by establishing beforehand a more precise and smooth division of time and speaking sections.

Timing: The presentation ran one minute over the mandatory duration, so the presenters must be careful not to exceed the strict 7:30-minute time limit. We recommend practicing the presentation multiple times beforehand; this will build greater knowledge of which sections to emphasize and which to potentially cut.

Certification: I/We certify that generative AI, incl. ChatGPT, has not been used to write this feedback. Using generative AI without permission is considered academic misconduct.

Riccardo Fragale (fragale@kth.se) & Ettore Mugisha Cirillo (emcir@kth.se)

hao8359:
Thank you for answering my question that I asked about whether a mutation test score is below 80 percent, do you just reject PR request or just send a warning to the engineering team but still accept it? The answer was clear that pr will be rejected even though the manually written checks seems fine.

PR #2965, opened 2026-09-02
Demo Proposal: Live CD Rollback on a To-Do List App, Comparing Manual vs. Automated Rollback

nalin431:
# Assignment Proposal

## Title

Live CD Rollback on a To-Do List App, Comparing Manual vs. Automated Rollback

## Names and KTH ID

- Nalin Kundu (nvkundu@kth.se)
- Hossein Ghadirzadeh (ghadirz@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

We'll use a small, standalone to-do list app (static frontend + a lightweight API) built specifically for this demo to demonstrate automated rollback mechanics within CD.

Our CI/CD pipeline (GitHub Actions) will deploy the app to [Render or a similar platform]. Our CI pipeline will then run tests before deployment, so broken code doesn't get merged. But some problems only show up once the change is merged to production (like a bad configuration or a dependency that isn't working). To catch these, an automated health check runs against the live app right after each deployment. During the demo, we'll ship a change that passes our CI gating tests but breaks the app once it's live. The health check catches this on its own, and the pipeline automatically rolls back to the last healthy release, with no manual intervention. We'll walk through the rollback strategy we chose and why we designed it to trigger automatically rather than being triggered by a developer.

**Relevance**

Rollback is a core DevOps practice: the ability to recover quickly and safely from a bad deployment is what makes frequent, automated releases viable. What we specifically want to highlight is the difference between a rollback a developer has to trigger manually and one the system performs on its own the moment it detects a problem. Running automatically is what actually makes continuous deployment safe to run unattended and at high frequency, which is the core promise of DevOps automation.

### DIFF
--- a/contributions/demo/week3/nvkundu-ghadirz/README.md
+++ b/contributions/demo/week3/nvkundu-ghadirz/README.md
@@ -1,0 +1,28 @@
+# Assignment Proposal
+
+## Title
+
+Live CD Rollback on a To-Do List App, Comparing Manual vs. Automated Rollback
+
+## Names and KTH ID
+
+- Nalin Kundu (nvkundu@kth.se)
+- Hossein Ghadirzadeh (ghadirz@kth.se)
+
+## Deadline
+
+- Week 3
+
+## Category
+
+- Demo
+
+## Description
+
+We'll use a small, standalone to-do list app (static frontend + a lightweight API) built specifically for this demo to demonstrate automated rollback mechanics within CD.
+
+Our CI/CD pipeline (GitHub Actions) deploys the app to [Render or a similar platform]. Right after each deployment, an automated health check runs against the live app. During the demo, we'll ship a deliberately broken change live (the health check detects the failure on its own), and the pipeline automatically rolls back to the last healthy release, with no manual intervention. We'll walk through the rollback strategy we chose and why we designed it to trigger automatically rather than being triggered by a developer.
+
+**Relevance**
+
+Rollback is a core DevOps practice: the ability to recover quickly and safely from a bad deployment is what makes frequent, automated releases viable. What we specifically want to highlight is the difference between a rollback a developer has to trigger manually and one the system performs on its own the moment it detects a problem. Running automatically is what actually makes continuous deployment safe to run unattended and at high frequency, which is the core promise of DevOps automation.
--- a/contributions/demo/week3/nvkundu-ghadirz/README.md
+++ b/contributions/demo/week3/nvkundu-ghadirz/README.md
@@ -21,7 +21,7 @@ Live CD Rollback on a To-Do List App, Comparing Manual vs. Automated Rollback
 
 We'll use a small, standalone to-do list app (static frontend + a lightweight API) built specifically for this demo to demonstrate automated rollback mechanics within CD.
 
-Our CI/CD pipeline (GitHub Actions) deploys the app to [Render or a similar platform]. Right after each deployment, an automated health check runs against the live app. During the demo, we'll ship a deliberately broken change live (the health check detects the failure on its own), and the pipeline automatically rolls back to the last healthy release, with no manual intervention. We'll walk through the rollback strategy we chose and why we designed it to trigger automatically rather than being triggered by a developer.
+Our CI/CD pipeline (GitHub Actions) will deploy the app to [Render or a similar platform]. Our CI pipeline will then run tests before deployment, so broken code doesn't get merged. But some problems only show up once the change is merged to production (like a bad configuration or a dependency that isn't working). To catch these, an automated health check runs against the live app right after each deployment. During the demo, we'll ship a change that passes our CI gating tests but breaks the app once it's live. The health check catches this on its own, and the pipeline automatically rolls back to the last healthy release, with no manual intervention. We'll walk through the rollback strategy we chose and why we designed it to trigger automatically rather than being triggered by a developer.
 
 **Relevance**
 


contributions/demo/week3/nvkundu-ghadirz/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Category
+
+- Demo
+
+## Description
+
+We'll use a small, standalone to-do list app (static frontend + a lightweight API) built specifically for this demo to demonstrate automated rollback mechanics within CD.
+
+Our CI/CD pipeline (GitHub Actions) deploys the app to [Render or a similar platform]. Right after each deployment, an automated health check runs against the live app. During the demo, we'll ship a deliberately broken change live (the health check detects the failure on its own), and the pipeline automatically rolls back to the last healthy release, with no manual intervention. We'll walk through the rollback strategy we chose and why we designed it to trigger automatically rather than being triggered by a developer.
```
algomaster99: I misunderstand two things here:
1. "health check" usually means a check which confirms that the server is alive (for example `ping`). How you use this term in the text seems like it is some form of tests.
2. Assuming health check does mean tests in this context, I feel it is strange to run the test _after_ deployment. They should be run before and reject deployment.

algomaster99 (changes requested):
The idea of rollback is good but I suggest to come up with a better scenario. See comments below.

contributions/demo/week3/nvkundu-ghadirz/README.md
nalin431: Thanks for the feedback! Yes, I think it was a little unclear, but we meant we will push a change that passses our CI tests that will still break the service and "kill" the server, which our demo will automatically rollback. Basically, we want to show how rollbacks are used for these changes that pass CI tests but still cause issues.

algomaster99 (approved):
Looks good now! Make sure that "healthcheck" is something which is hard to incorporate into test suites. Otherwise, there wouldn't be a need for a rollback.

h4lan:
## Feedback on "Live CD Rollback on a To-Do List App, Comparing Manual vs. Automated Rollback"

### Understanding:

The demo shows an automated rollback mechanism in a CD pipeline GitHub Actions deploys a To-Do app to Render, post deploy a health check detects a broken live release that passed CI, and the pipeline automatically rolls back to the last healthy version without human intervention.

### Technical Depth

In terms of technical depth, the presentation looks strong on paper. students mentioned CI gating, a separate CD deploy step, a live health check, and an automated rollback trigger, which is genuinely multiple interacting components.

Maybe some of those terms could have been explained, allowing the students to show their understanding and also lengthening the presentation.

### Liveness

During the presentation, the students actively interacted with the code and GitHub actions, showing their designed workflow in real time.

Since there are blanks during a demonstration while waiting for a result to show on the screen, having one of the students give general comments by justifying using certain technologies answering Why did we use Render? or choosing this specific rollback method over other options like blue-green, just redeploying the last artifact directly, can help retain the audience's attention.

### System Reasoning

During the demo, the students explained clearly how the components interact with each other where the health check lives, what signal triggers the rollback and how the automatic rollback is happening. They have also included a flow diagram on their slides before diving into the code and a slide that explains how the rollbacks minimize the production downtime.

### Relevance

The presentation explains that the ability to recover quickly and safely from a bad deployment is what makes frequent, automated releases viable.

It would be great to show the difference between a rollback a developer has to trigger manually and one the system performs on its own the moment it detects a problem, as this was mentioned in the pull request. 

### Design Decisions

The students justified the rollback strategy and the design decisions for the demo and why the rollback is automatic rather than manual.

As a conclusion, they should make sure the demo actually states the alternatives: blue-green, rolling restart, and re-deploying previous artifacts. It would also be great to justify why some tools (like Render) were used.

### Reflection

The students clearly stated the types of errors that CI doesn't detect but automated rollback prevents from affecting production.

They did not mention any existing limits or drawbacks of their method, nor did they provide alternatives to help in fixing those. For example, a system could potentially rollback to a version with known vulnerability issues. Setting up feature flags (or feature "toggles") and canary deployment practices could respectively allow teams to disable the breaking functionalities only or could limit the impact that a faulty newer version can have.

### Narrative

The students started by clearly presenting the problem, and exposing their solution and presenting the system and the tool used to implement the rollback. Then, the demonstration shows the problem taking place, and how it was fixed. This structure made it easy to follow what the demo was about.

However, the implications of automated rollback were not explicitly stated. The presentation included everything that was mentioned in the pull request but the "difference between a rollback that the developer has to trigger manually and one the system performs on its own." Even if the demonstration could not include this comparison, mentioning this comparison on the slides could have completed the structure and would have made for a great closing statement. A way to mention the importance of automated rollback could be: "It ensures that continuous deployment is a viable practice and that merging new functionalities frequently doesn't break production environments as frequently."

### Speech

With clear and understandable speech, both students showed that they were confident about their knowledge, both during the mock presentation and when answering our questions. Their answers to our questions were clear.

For the actual demonstration on the computer, preparing a script could help, since finding the right words to describe what is being shown while executing the demonstration can be challenging.

### Timing

The mock presentation lasted around 6 minutes, which is a few seconds short of the required time. Yet, there were some elements that were skipped over. This could easily be improved if the students mention the missing points raised beforehand. For instance, comparing the practice of rolling back automatically with manual rollback and describing the .yaml files for the jobs.

Conceptually this is a strong, focused demo idea with clear DevOps relevance. The two areas to prioritize before delivery are making the live interaction more lively and adding an explicit reflection on the limitations of automated rollback.


## Certification

Sangeetha Murugesan (sanmur@kth.se) and Halan Ouensanga (halano@kth.se)

h4lan:
Me and @sangeetha-murugesan-sda9 met with @nalin431 and @hosseinghzadeh so they could present their demo to us: the previous comment shows the feedback we gave after the presentation.

PR #3008, opened 2026-09-10
Demo Proposal: Solving configuration drifts using ArgoCD's automated self-healing feature 

Gabjea:
# Assignment Proposal

## Title

Solving configuration drifts using ArgoCD's automated self-healing feature 

## Names and KTH ID

  - Alexandru Gânju (ganju@kth.se)
  - Robert Fedus (fedus@kth.se)

## Deadline

Week 5

## Category

Demo

## Description

We're going to present a simple scalable microservice deployed on a live Kubernetes cluster.
In a ConfigMap we'll have a feature flag for our app and showcase how drifts happen by manually making changes to some resources in the live cluster.
Then, we'll show how to use the auto-sync and self-healing features and how ArgoCD solves the drift by syncing from our git repository.
In the end, in the case of using a Horizontal Pod Autoscaler (HPA), ArgoCD will fight against it when the above features are enabled. We'll explain how and when you might want to ignore some drifts.

**Relevance**

From a DevOps perspective it is really important to have a maintainable and auditable infrastructure.
We chose ArgoCD because of its core features that enables DevOps Engineers to achive that by using GitOps principles (have the git repository act as the source of truth).

### DIFF
--- a/contributions/demo/week5/ganju-fedus/README.md
+++ b/contributions/demo/week5/ganju-fedus/README.md
@@ -1,0 +1,31 @@
+
+# Assignment Proposal
+
+## Title
+
+Solving configuration drifts using ArgoCD's automated self-healing feature 
+
+## Names and KTH ID
+
+  - Alexandru Gânju (ganju@kth.se)
+  - Robert Fedus (fedus@kth.se)
+
+## Deadline
+
+Week 4
+
+## Category
+
+Demo
+
+## Description
+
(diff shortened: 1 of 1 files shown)


ericcornelissen (changes requested):
The proposal looks good but the chosen week seems irrelevant to the topic, please pick another week. Also, please spell out HPA.

Gabjea:
Thank you! I've updated the proposal!

ericcornelissen (approved):

PR #2978, opened 2026-09-03
Safe Continuous Deployment with Blue-Green Deployment

TheArtemis:
# Assignment Proposal

## Title

Safe Continuous Deployment with Blue-Green Deployment

## Names and KTH ID

- Lorenzo Deflorian (ldef@kth.se)
- Riccardo Fragale (fragale@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

Blue-green deployment is a release strategy that uses two identical production environments: one active environment serving users and one inactive environment. A new application version is deployed to the inactive environment and tested before production traffic is redirected to it. If the checks fail, the active environment continues serving traffic. If problems appear after the switch, traffic can quickly be redirected to the previous environment.

During the demo, deployed in [Killercoda](https://killercoda.com/) we will show a simple Continuous Deployment pipeline that:

1. Starts with the current application version running in the active environment.
2. Deploys a new version to the inactive environment.
3. Runs automated health checks against the new version.
4. Switches traffic to the new environment when the checks pass.
5. Demonstrates a failed deployment where health checks prevent a faulty version from receiving traffic.
6. Demonstrates a rollback by redirecting traffic to the previous working environment.

This will allow us to compare successful and unsuccessful deployments and show how blue-green deployment can reduce downtime and deployment risk.

**Relevance**

Continuous Deployment is an important DevOps practice for releasing changes quickly and safely. This demo illustrates how automated validation, failure handling, traffic switching, and rollback can reduce production downtime and prevent faulty releases from reaching users.

PR #3023, opened 2026-09-14
Demo proposal: Dependency-Aware Blue/Green Continuous Deployment with GitHub Actions

johnsward:
# Assignment Proposal

## Title

Demo proposal - Dependency-Aware Blue/Green Continuous Deployment with GitHub Actions

## Names and KTH ID

  - John Swärd (johnsw@kth.se)
  - Dimitris Bakalis (bakalis@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

We want to demonstrate a Continuous Deployment pipeline for three dependent Spring Boot microservices (`users`, `auth`, `orders`, where `auth` depends on `users` and `orders` depends on `auth`), built on GitHub Actions with a
self-hosted runner. On every push, the pipeline detects which services changed, resolves their declared dependency graph to find every impacted downstream service, and deploys each one in order using blue/green deployment: a new version is started in a second, idle slot alongside the currently serving one, smoke-tested directly, and only switched into live traffic via nginx once it passes. If a service's deployment fails, its previous slot keeps serving traffic untouched, and every service depending on it is automatically blocked and skipped rather than deployed on top of a
broken dependency. This is built with **GitHub Actions** for CI/CD orchestration, a **self-hosted runner** so the pipeline can deploy to our own machine, **Maven** and **Docker** to build each service, and **nginx** as the single component that holds stable ports and performs the zero-downtime traffic switch between blue/green slots.

**Relevance**

This demo reflects core DevOps principles by showing how a change to one service in a dependency chain can be safely and automatically propagated through its dependents, with each deployment gated by its own health check via zero-downtime blue/green deployment rather than a single all-or-nothing rollout. This approach lets teams reduce the blast radius of a faulty deployment to just the services that actually depend on it, without manual intervention.


algomaster99:
Missing student registration :jsward

ericcornelissen (approved):
Nice proposal. Since #2978 already covers Blue/Green deployments, please focus your demo more on the other aspects of the proposal.

PR #3021, opened 2026-09-14
Add demo proposal: Blue-Green Deployment with Automated CI/CD

Septimusmi:
# Assignment Proposal

## Title

Blue-Green Deployment with Automated CI/CD for a Node.js Demo App

## Names and KTH ID

  - Miami Alvelistin (miami@kth.se)
  - Albin Wallenius Woxnerud (alww@kth.se)

## Deadline

Week 3

## Category

Demo

## Description

Our demo shows how to turn a manually run web app into a self-updating local deployment that pushes new code automatically with no downtime.

We fork the open-source nodejs-demoapp (a small Express app) and build a GitHub Actions pipeline around it. On every push, the pipeline lints the code, runs integration tests, and builds a Docker image. When a change lands on main, a self-hosted runner on our local machine triggers a blue-green deployment. It builds a new container and health-checks it against a readiness endpoint. Traffic switches over only once the container is confirmed healthy. The old container is then removed. If the health check fails, the previous version keeps serving traffic, and the deployment aborts automatically.

During the live demo, we will push a code change to the repository and show the full chain reacting in real time: the pipeline running its checks, the new image being built, and the running application being hot-swapped for the new version without any observable downtime or manual intervention. We will also intentionally push a broken change to demonstrate that the pipeline blocks a bad deployment before it ever reaches the running container.

**Relevance**

This demo covers several core DevOps practices: continuous integration (automated linting and testing on every commit), continuous deployment (an automated, hands-off path from commit to running software), and zero-downtime release strategies (blue-green deployment). Automated quality gates act as a safety net, while health checks and gradual traffic cutover make deployments safe and reversible. It also covers a common constraint: deploying to infrastructure a cloud-hosted CI runner can't reach. A self-hosted runner bridges that gap, a pattern used for on-premises deployments.

PR #3029, opened 2026-09-16
Demo proposal: Preventing Version Skew Between Independently Deployed Services

albinwwoxnerud:
# Assignment Proposal

## Title

Preventing Version Skew Between Independently Deployed Services

## Names and KTH ID

- Miami Alvelistin (miami@kth.se)
- Albin Wallenius Woxnerud (alww@kth.se)

## Deadline

- Week 3

## Category

- Demo

## Description

Two services in separate repositories can both pass CI and still break when deployed together. Each pipeline tests only its own service, so the exact production combination may never be tested.

We demonstrate this with a provider and consumer over HTTP. Both have independent GitHub Actions pipelines that test, build, push and deploy.

First, both services receive matching API changes. Both pipelines pass, but deployment order briefly creates an incompatible pair and the app breaks.

Then we fix it by pinning both image digests in one release manifest, testing that exact pair and deploying them together. Only tested combinations reach production.

**Relevance**

Continuous Deployment is often described one service at a time, but real systems depend on services staying compatible. A green pipeline proves one service works, not that the deployed combination does.

We also cover the limits of coordinated releases. Rollouts can still create brief version mismatches, so breaking contract changes should remain backward compatible. Coupling releases also reduces service independence, and our local deployment uses a trusted self-hosted runner.


albinwwoxnerud:
Topic change from https://github.com/KTH/devops-course/pull/3021

contributions/demo/week3/miami-alww/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Week 3
+
+## Category
+
+- Demo
+
+## Description
+
+We will demonstrate a continuous deployment pipeline that prevents older GitHub Actions runs from overwriting newer releases. Each push tests the application, publishes a commit-tagged Docker image to GitHub Container Registry, and deploys it through a self-hosted runner to a local k3d Kubernetes cluster. Stale commits are rejected before deployment.
```
algomaster99: It is not clear to me what do you mean by stale commit. Sometimes you genuinely want to rollback to a previous version (stale commit?) because the newer commits introduced some regression.

contributions/demo/week3/miami-alww/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+During the demo, we will deploy a new version, attempt to deploy an older version, and delete a running pod. Kubernetes will recreate the pod while another replica continues serving traffic.
+
+**Relevance**
+
+CI/CD pipelines can finish out of order, causing an older release to replace a newer one. This demo combines release-order protection with Kubernetes reconciliation to show automated, reproducible deployment and recovery. We will also discuss the limitations of a local cluster and trusted self-hosted runner.
```
algomaster99: > CI/CD pipelines can finish out of order, causing an older release to replace a newer one.

Interesting. In what scenario would you do this? Why is the older version even being deployed if there is a CD running for a newer version?

algomaster99 (changes requested):
Thanks! Looks very interesting. Just want you to clairfy the use case a bit and then we are good.

Also, I guess #3021 is outdated now. Can you close it?

contributions/demo/week3/miami-alww/README.md
albinwwoxnerud: "Stale commit" was a very bad choice of words admittedly.

What we meant was a pipeline run for an older commit that finishes after a newer commit has already been deployed.

We did not mean that older versions should be blocked. Deploying an older version can be exactly what you want in a rollback, and that should be an explicit action the pipeline allows.

contributions/demo/week3/miami-alww/README.md
albinwwoxnerud: The scenario came from something we actually hit, but it was not a single-repository problem.

We had two services in separate repositories, each with its own pipeline, with a dependency between them. Both were pushed at roughly the same time. One started building first but finished later, so we ended up deploying a combination of artifacts that had never been tested together.

Your question is fair. Within a single repository with one CD pipeline, this is much harder to justify. `concurrency` with `cancel-in-progress` covers most cases. What remains, like manually re-running an old workflow or retrying a failed job after a newer release has landed, is fairly narrow.

What we actually hit was a cross-repository artifact coordination problem. We presented it as a deployment ordering problem, which made the proposal misleading.

algomaster99: > We had two services in separate repositories, each with its own pipeline, with a dependency between them. Both were pushed at roughly the same time. One started building first but finished later, so we ended up deploying a combination of artifacts that had never been tested together.

Ah so you mean that one service was up to date while the other was stale? This is nice for a demo. Does the demo ensure that these two releases are deployed at the same time?

albinwwoxnerud: Not simultaneously, no.

Strict simultaneity is not really possible. Any rollout has a window where one service is updated before the other.

What we can guarantee is that the exact combination reaching production was tested together. The services still build independently, but their image digests are pinned in a release manifest, integration-tested as a pair and deployed as one release. Neither service deploys independently.

We’ll first show the naive setup, where two pipelines deploy on push and can produce a combination that was never tested. Then we’ll show the manifest-based setup, where that cannot happen.

There is still a brief mismatch during rollout. The production solution is backward-compatible contract changes. We’ll cover that part, but leave it out of the demo.

albinwwoxnerud: I have updated the proposal to more accurately reflect this. Thank you for your time.

algomaster99 (approved):
Clearer to me now! Thanks!

PR #3038, opened 2026-09-17
Executable tutorial submission: LLMOps prompt regression testing

JingzeGuo:
# Assignment Proposal

## Title

Prompt regression testing for LLM applications (LLMOps)

## Names and KTH ID

  - Jiaxun Wei (jiaxun@kth.se)
  - Jingze Guo (jingze@kth.se)

## Deadline

- Task 2

## Category

- Executable tutorial

## Description

**Tutorial:** [Google Colab](https://drive.google.com/file/d/1-0P0LdPLC0fstHJVhJ_B3xGATXP8OIYL/view?usp=sharing)

We want to write a Google Colab notebook that shows how to keep an LLM application stable when its prompt or its model changes. The reader writes a promptfoo configuration with test cases and assertions, runs it against a small instruction-tuned model that is loaded locally in the notebook. The reader then breaks the prompt on purpose, watches the assertions fail, and finally wires the evaluation into a pipeline, so that a failed evaluation blocks the change.

**Relevance**

An LLM application is defined as much by its prompts and its model version as by its code, and both change often. Without evaluation, a reworded prompt or a model upgrade can silently degrade behaviour, which is the LLMOps version of a regression. The tutorial makes that behaviour testable and automated, so that changes to a model-driven system get the same review discipline as changes to code.


algomaster99:
We will evaluate this when we are done with all the seminars. Thanks for sharing the link!

